PR #37, opened 2017-06-07
Fixes badly formatted channel names

Cyberboss:


TGServerService/Chat.cs
```diff
@@ -308,6 +308,19 @@ public string SetProviderInfo(TGChatSetupInfo info)
 			}
 		}
 
+		//trims and adds the leading #
+		string SanitizeChannelName(string working)
+		{
+			if (working == null)
```
AnturK: string.isNullOrWhitespace instead of splitting this around.

PR #39, opened 2017-06-09
Github issue reporting interface

Cyberboss:
Forum thread: https://tgstation13.org/phpBB/viewtopic.php?f=5&t=11648&p=305814
Sister PR: https://github.com/tgstation/tgstation/pull/28288
See the readme update


MrStonedOne:
This shouldn't honestly be done here, the tgstation13 website would be the proper place for such an interface.

If needed, I could create an endpoint that doesn't require ssl to allow for such a thing to be a in game interface that exports to a php script.

Or we could create a dll wrapper for libcurl to allow for this to be done in the native github api. (or use the ps version of curl)

But I do not feel it to be productive or sane to have an api to access an api, nor do i see any point in shoe horning this into the server toolkit

PR #46, opened 2017-06-10
Service is now started only by the installer

Cyberboss:
Nothing special


MrStonedOne:
Explain.

What do you mean by only, what was also starting it before?

Cyberboss:
Windows uses installutil to install services. You can write c# code to hook
into the process. I wrote code to start the service. the installer program
also does this itself. This is just removing some redundancy

PR #54, opened 2017-06-12
DreamDaemon is now suspended while symswap happens

Cyberboss:
Oh wait, I forgot we had TOTAL CONTROL!!!


Cyberboss:
DNM while this comment is here, need to test

PR #57, opened 2017-06-12
Logging Refactor + Event IDs

### LINKED ISSUE #48: Commands to run on certain actions
I'm in the middle of building a docker container framework that can be used to host "edge" cdn byond servers that send the user the resources and then send them on to the real server.

The last step would be some way to know when the server's rsc has changed and trigger a rebuild of the docker image (I'll have it grab the rsc via the smb share that already exists for the game folder, that part isn't hard)

So I was thinking, it would be handy to create events that ran things.

You could do this the user friendly way and make an interface for the different types of events (restart, round end, crash, update, compile, etc), or you can just push unique event ids in windows event viewer and i can just make scheduled tasks based off of them.

Just try to be as granular as sanely possible and very particular in your documentation about what events trigger when.

Cyberboss:
@MrStonedOne  @JamieH The list: https://github.com/tgstation/tgstation-server-tools/pull/57/files#diff-e9bc64778053ef4bd86682c5bdc3e08fR14

Closes #48


Cyberboss:
I've multiplied event ID's by 1000, the plan is to use the 1000x level to denoted the event type and the 1-999 level to denote which server instance it occurred on when I implement instanced servers

JamieH:
@Cyberboss That sounds bad, can you not just push the instance ID as additional metadata in the event?

Cyberboss:
The data associated with an event is the ID, the service, the type (warning, error, info, etc...) and the text. @MrStonedOne what's your take?

MrStonedOne:
ya, it seems doing different event ids is the only way to let you customize the tasks for each server other than running different services

Unless you register task keywords, but i don't know how that works

PR #69, opened 2017-06-21
Adds -clean to dm compile options

Cyberboss:
No idea what this does but tgs2 does it


MrStonedOne:
You should still really delete everything in the staging folder before copying from the repo.

Cyberboss:
That already happens

PR #73, opened 2017-06-22
Adds a version command to chat bot

Cyberboss:


Cyberboss:
I have something important to do for sybil and this is in the way, so speedmerging

PR #77, opened 2017-06-23
Fixes Commits not happening for modifications and deletions

Cyberboss:


Cyberboss:
Speedmerging because fuck this bug

PR #78, opened 2017-06-23
Fix bugs relating to doing multiple operations during the same round

### LINKED ISSUE #64: RSC corruption likely caused by an old rsc sticking around in the staging folder.
If there is a rsc in the folder, byond will try to re-use it, and even use it for resources not in the actual project folder.

This (feature) of byond is horribly bugged, random bits of old versions of now not used files will get interspaced where nulls or rsc metadata should be.

Example:

`Ó....x">Photocopied papers will now retain their stamps</li>....` (. = unprintable character)
The Ó is the exact spot the last file ended (an ogg) at this point, there should be 18 bytes of meta data, followed by 4 bytes of payload size, a null terminated file name string (not part of the payload size), and the payload

Instead we have a random changelog entry.

This is a real example, from hippie.

tgs2 deleted the old staging folder every time before copying over from the repo to prevent things like this (and prevent now deleted files sticking around)

MrStonedOne:
Look ma! I'm working on c# code in the web editor!

I don't know what calls this code, there might be cases where we don't want this to unstage the update, (ie, its only checking, but has no plans to do anything)

Fixes #64

### DIFF
--- a/TGServerService/Compiler.cs
+++ b/TGServerService/Compiler.cs
@@ -264,6 +264,9 @@ namespace TGServerService
 				}
 				catch	//held open by byond
 				{
+					//This means there is a staged update waiting to be applied, we have to unstage it before we can work
+					Directory.Delete(GameDirLive);
+					CreateSymlink(GameDirLive, TheDir);
 					return InvertDirectory(TheDir);
 				}
 			}


Cyberboss:
Probably fixes #64

MrStonedOne:
The only other way (and might be the most proper way) would be to do a/b/c and cycle them, then this issue could never be an issue if we make sure the folder is neither staged or active. (3 states, folders can have, staged (update is staged here), active (dd is running from this folder), inactive (none of the above))

Issue #80, opened 2017-06-26
Headmins add IRC alerts on the phrase "watchdog", but the new toolkit uses this phrase excessively


MrStonedOne:
Having these alerts trigger on 4 channels is annoying, having these alerts trigger on random shit like the toolkit updating is annoying.

There needs to either be a massive scaling back of the times we use this word, or a new word we can bind to has to be created, for headmins who want to get alerts when the server crashes/restarts.

PR #79, opened 2017-06-26
Chat Refactor

Cyberboss:
Allows multiple chat providers at once
Configurable channels for dev/admin/game/watchdog messages
Allows IRC modes to be used for identification
Allows Discord roles to be used for identification

Fixes #26 
Fixes #75 
Fixes #80

PR #81, opened 2017-06-26
Changes some chat messages

MrStonedOne:
watchdog != watch dog

This doesn't address #80 fully because it doesn't solve the issue that these messages go out to all channels.

Sadly the only fix to that would be to make message types and define what channels each message type goes out to.


Cyberboss:
#79 is working on the big things

MrStonedOne:
ya, you should give an actual PR description on that pr does and changes.

PR #85, opened 2017-06-28
Fixes a potential race condition when reading the Interop port

Cyberboss:


Cyberboss:
No, this config needs to go

PR #91, opened 2017-06-29
Chat command refactor

Cyberboss:
Adds help text and the revision command
untitled


MrStonedOne:
What does it say when you hover over the (i)?

PR #118, opened 2017-07-05
Removes an unused var

Cyberboss:


Cyberboss:
Ban me I'm seizing control

PR #132, opened 2017-07-11
Changelogs will no longer be push if the current commit does not match the tracked remote commit

Cyberboss:


RemieRichards:
you're no longer generating changelogs for testmerged content.

MrStonedOne:
ya, you still have to generate changelogs on test merges, just not push them.

and if you do update and test merge, you have to generate twice, pushing the first one.

RemieRichards:
yeah, that's what my PR did, but I didn't have the commit matching checks and I got annoyed at HippieGordonRamsay so I closed it.

MrStonedOne:
part of the issue is that you put into one function, (with ifs and args) what should be 5 seperate functions.

cyberboss, you have one function that calls 5 seperate functions based on args given by five seperate bits of code, and thats a lame way to do it. 

`UpdateServer()` needs to go, the entire function needs to be killed.

The switch tree on `fuAction` should be where that code is.

You could enable this by removing some of the boilerplate.

There is also no need to check the origin master if you only push changelogs after a reset operation, because then the repo is already following origin master.

The switch tree should look like this:

````c#
			switch (fuAction)
			{
				case FullUpdateAction.Testmerge:
					TestMergePr();
					CompileChangelog();
					CopyFromGit();
					Compile();
					Stage();
					break;
				case FullUpdateAction.UpdateHard:
					ResetUpdateGit();
					CompileChangelog();
					PushChangelog();
					CopyFromGit();
					Compile();
					Stage();
					break;
				case FullUpdateAction.UpdateHardTestmerge:
					ResetUpdateGit();
					CompileChangelog();
					PushChangelog();
					TestMergePr();
					CompileChangelog(); //this is correct
					CopyFromGit();
					Compile();
					Stage();
					break;
				case FullUpdateAction.UpdateMerge:
					MergeUpdateGit();
					CompileChangelog();
					CopyFromGit();
					Compile();
					Stage();
					break;
				case FullUpdateAction.Reset:
					ResetGit();
					CompileChangelog();
					CopyFromGit();
					Compile();
					Stage();
					break;
			}
````

This is foolproof, and its much much much more easier to understand, visualize the flow, and ensure everything is being sane.

MrStonedOne:
note, you also have another option here.

This isn't batch code anymore, storing arrays and lists of things is a trivial task now. we can just use the list of active test merges.

reset all test merges on every update operation, and just re-test merge them, in order they were added.

if you made that proper seperate test merge gui to add/remove test merges without actually test merging them, then click a button to apply them, this would have all naturally fell into place.

The flow should be that the user lists the pr numbers they want test merged, then, they click a button to test merge them, and the code has to handle merge errors sanely. (attempt to abort the merge, if the abort fails, reset the repo, then ask the user if they want to remove that pr from the list and retry or abort)

Once this kind of system is in place, we can do things like just removing a test merge, and make updates always reset first, then re-test merge the active test merges.

Shadowlight213:
Just do mso's foolproof simple solution. There is zero reason to be pushing anything at all if it's not a hard reset to master. No need for any fancy tracking, just make it only run on hard reset update, and hard reset update and testmerge.

PR #134, opened 2017-07-14
Removes ITGServerUpdater. Locks changelog pushing to when local HEAD matches remote

Cyberboss:


TGServerService/Repository.cs
```diff
@@ -671,7 +671,30 @@ public void SetCommitterEmail(string newEmail)
...
+		{
+			lock (RepoLock)
+			{
+				if (LoadRepo() != null)
+					return false;
+				var R = Repo.Network.Remotes["origin"];
+				try
+				{
+					Commands.Fetch(Repo, R.Name, R.FetchRefSpecs.Select(X => X.Specification), null, null);
+					return Repo.Head.IsTracking && Repo.Head.TrackedBranch.Tip.Sha == Repo.Head.Tip.Sha;
```
Cyberboss: I still want to keep this check because it can only guarantee that a certain fuck up doesn't happen again

TGControlPanel/ServerPage.cs
```diff
@@ -373,25 +373,68 @@ private void ServerGRestartButton_Click(object sender, EventArgs e)
...
-					updateError = Updater.UpdateServer(TGRepoUpdateMethod.None, false, (ushort)testmergePR);
+					updateError = Repo.MergePullRequest(testmergePR);
+					updateError = Repo.GenerateChangelog(out string error2);
+					if (updateError == null)
+						updateError = DM.Compile(true) ? null : "Compilation failed!";
 					break;
 				case FullUpdateAction.UpdateHard:
-					updateError = Updater.UpdateServer(TGRepoUpdateMethod.Hard, true);
+					updateError = Repo.Update(true);
+					updateError = Repo.GenerateChangelog(out error2);
```
MrStonedOne: You shouldn't be stopping compiling if the change log fails to generate, its an optional thing, having python installed should be an optional thing.

PR #128, opened 2017-07-07
Fixes Interop

Cyberboss:

PR #139, opened 2017-07-27
Chages relay API from shell() and sockets to call() and the interface DLL

### LINKED ISSUE #83: Interop in either direction is non-functional after updating the service
Only a hard restart from the server tools remedies this

### LINKED ISSUE #104: Server stops relaying to irc or discord randomly
The bot may disconnect from irc or just start ignoring discord, you have to disable and re-enable the chat interface to fix it.

Cyberboss:
By adding a C export to the interface DLL that BYOND can use

Closes #128 and Closes #83 and Closes #104 due to not using flaky sockets anymore

Sister PR https://github.com/tgstation/tgstation/pull/30507


Cyberboss:
Potential issue: How hard will it be to replace the dll during upgrades due to byond being a butt. Could use the same system updating byond does to upgrade it

MrStonedOne:
byond should be closing the dll during round restarts.

if not, we could force the issue within the dll by forcing an unload (i can't remember how, but i heard/read something about that)

Cyberboss:
Granted, there's no need to copy it every time anyway because this API should NEVER EVER change

PR #143, opened 2017-08-26
Add an update and restart command to the chat bot

JamieH:
Ignore the branch name, originally I was going to try and add GRPC to the service so I could communicate with it over the Internet from a non .NET applicatiuon but wasn't really sure the best way to implement this.

It'd be great if you could perhaps expose the ServiceInterface over SOAP @Cyberboss

### DIFF
--- a/TGServerService/ChatCommands.cs
+++ b/TGServerService/ChatCommands.cs
@@ -41,7 +41,20 @@ namespace TGServerService
 		public RootChatCommand()
 		{
 			PrintHelpList = true;
-			Children = new Command[] { new CheckCommand(), new StatusCommand(), new PRsCommand(), new VersionCommand(), new AHelpCommand(), new NameCheckCommand(), new RevisionCommand(), new AdminWhoCommand(), new ByondCommand(), new KekCommand(), new RelayRestartCommand() };
+			Children = new Command[] { new CheckCommand(),
+                new StatusCommand(),
+                new PRsCommand(),
+                new VersionCommand(),
+                new AHelpCommand(),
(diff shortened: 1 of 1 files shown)


MrStonedOne:
This needs to be configurable/disablable.

\>be me
\>find admin level RCE
\>put fix pr up
\>test merge fix pr
\>admin wants to exploit rce
\>commands the server to un-test merge over irc by requesting it hard update
\>exploit is back on the server

Rareish, but a genuine concern.

\>be me
\>update server
\>everything breaks
\>roll back server
\>nag coderbus
\>admin comes on line
\>sees a pr they liked/made got merged but isn't live
\>updates server via irc command
\>everything breaks
\>can't fix it because thats not a command they can use

It's much easier to keep the few people with remote on the same page on when upstream is broken then it would be for all admins

\>be me
\>find major bug
\>put fix pr up
\>test merge fix pr
\>admin comes on line
\>sees a pr they liked/made got merged but isn't live
\>updates server via irc command
\>merge conflict with some test merge
\>hard updates the server
\>bug is back
\>can't fix it because test merge isn't a command


\>add test merge command
\>now every admin can create their own RCE exploit by making a malicious pr and then test merging it with the irc command

***

I don't want to keep hippie from having this feature (but i'd strongly suggest against it) but I do want to be able to disable it for /tg/

Jordie0608:
Can corroborate there have been a couple of times we had to block updates entirely to prevent certain scrubs from updating without looking at the warnings not to. Introducing the ability to more would only exacerbate.

PR #148, opened 2017-09-01
Implement server chat commands

Cyberboss:
Sister PR with deets: https://github.com/tgstation/tgstation/pull/30343


TGServerService/Interop.cs
```diff
@@ -118,7 +128,7 @@ string SendTopic(string topicdata, ushort port)
 						string returnedString = "NULL";
 						try
 						{
-							var returnedData = new byte[512];
+							var returnedData = new byte[UInt16.MaxValue];
```
Cyberboss: Needful

PR #149, opened 2017-09-06
Adds submodule update handling

Cyberboss:
Needed for https://github.com/tgstation/tgstation/pull/30457


Cyberboss:
Funnily enough I had already handled this during cloning lol

Cyberboss:
Upstream bug triggered when checking out a commit with a submodule (first init), then back to an older commit (deinit), then back to submodule commit and trying to update (bugged) https://github.com/libgit2/libgit2/issues/3820

PR #156, opened 2017-09-08
Adds HTTPS interface and makes the GUI/CLI work remotely

Cyberboss:
@JamieH, so I'm got high and suddenly it was working.

TL;DR Send your windows login over this internet through HTTPS and it will block access to the commands unless the authentication is correct.

There's a bit of fuckery to bind the certificate to the port which will be hardcoded to (38607 mainly because I'm lazy and I just want this working for now): Instructions here: https://docs.microsoft.com/en-us/dotnet/framework/wcf/feature-details/how-to-configure-a-port-with-an-ssl-certificate

WIP: the cli/gui don't actually use this stuff yet. But TL;DR everything works the same. First call SetRemoteLoginInformation() to set the address, username, and password, then call VerifyConnection() to see if you're connected, then call Authenticate() to see if you got it right. After that the interface works like normal. I need to do some optimization to cut down on GetComponent() calls, which are expensive over the internet but I'll do that soonish.

It works with any windows user login information on the computer the service is running on. Accounts without a passwoord won't work

This shit might actually be done for 3.1.


JamieH:
Holy shit.

MrStonedOne:
>It works with any windows user login information on the computer the service is running on. Accounts without a passwoord won't work

I do need a way to either whitelist or blacklist users and/or groups.

There are a few utility accounts in use that are for automatic transferring of files around the vms and servers, and I generally try to lock these accounts down so they can't be used for anything but read only file transferring of the very limited amount of files they are given access to.

I would normally just suggest you check that the user can login interactively and/or check they had remote login privileges. But at some point I plan to restricting direct remote access to more technology minded people and giving non-techy headmins only remote control panel access (once I feel its prime time ready), so ideally we would want a more future proof design for that.

Likely aiming for some way to specify a usergroup that could be whitelisted and/or blacklisted.

Jordie0608:
Citrix you say ( ͡° ͜ʖ ͡°)

Cyberboss:
@MrStonedOne You can now user the command line to specify which user group has access to the service. This also limits access locally! Only administrators have guaranteed local access now.

Command line can connect to remote services using the `remote` command in interactive mode or by adding `-c username:password@address` to a one shot command.

The control panel has a new login page, address and username are saved in the config, password isn't.

untitled

All in all good to go I say!

Cyberboss:
Reminder to fix the deployment script before I merge

MoreRobustThanYou:
Add the "limited accounts" please

Cyberboss:
later

PR #157, opened 2017-09-09
	Adds a preaction framework so we can run a .bat file before or after compiling the code

JamieH:
We need a way to do shit before and after the code is compiled but we need a blocking way of doing this, so we can't use the events, this allows us to do that.


TGServerService/Compiler.cs
```diff
@@ -383,6 +383,13 @@ void CompileImpl()
 					}
 				}
 
+                if (!PrecompileHook())
```
Cyberboss: Tabs sir

TGServerService/PreactionHandler.cs
```diff
@@ -0,0 +1,65 @@
...
+using System.IO;
+
+namespace TGServerService
+{
+    // Some useful functions for triggering pre action events
+    partial class TGStationServer
+    {
+        const string EventFolder = "EventHandlers/";
+
+        string GetPath(string eventName)
```
Cyberboss: Instead of using strings you should use the event ids and call the handlers from the logging function

TGServerService/ServerService.cs
```diff
@@ -76,6 +76,8 @@ public enum EventID
 			ChatBroadcastFail = 6400,
 			IRCLogModes = 6500,
 			SubmoduleReclone = 6600,
+			PreactionEvent = 6700,
+            PreactionFail = 6800
```
Cyberboss: teb

TGServerService/PreactionHandler.cs
```diff
@@ -0,0 +1,65 @@
+﻿using System;
+using System.Diagnostics;
+using System.IO;
+
+namespace TGServerService
+{
+    // Some useful functions for triggering pre action events
```
Cyberboss: Check tabs here too

TGServerService/PreactionHandler.cs
Cyberboss: e.g. You could have a file titled `35xx - Do something after repo cloned.bat`

JamieH:
I fixed tabs/spaces. The only problem with using the event logger function to do this is I want it to fail during compile if the bat file doesn't return the correct exit code.

Version.cs
```diff
@@ -12,6 +12,6 @@
 // [assembly: AssemblyVersion("1.0.*")]
 
 //It's impossible to make these a define, don't say I didn't warn you 
-[assembly: AssemblyVersion("3.0.90.2")]
-[assembly: AssemblyFileVersion("3.0.90.2")]
-[assembly: AssemblyInformationalVersion("3.0.90.2")]
+[assembly: AssemblyVersion("3.0.90.4")]
+[assembly: AssemblyFileVersion("3.0.90.4")]
+[assembly: AssemblyInformationalVersion("3.0.90.4")]
```
Cyberboss: I'll handle the version bump in it's own commit for the release

README.md
```diff
@@ -89,6 +89,9 @@ The process to update the server while DD is running is a functioning work in pr
 * `RepoKey/`
 	* This contains ssh key information for automatic changelog pushing. 
 
+* `EventHandlers/`
+	* This contains batch files that run after certain events, currently only precompile and postcompile events are implemented.
```
Cyberboss: You should better describe what the .bats should be titled

JamieH:
cc @Cyberboss

TGServerService/Compiler.cs
```diff
@@ -383,6 +383,13 @@ void CompileImpl()
 					}
 				}
 
+				if (!PrecompileHook())
+				{
+					lastCompilerError = "The precompile hook failed";
+					compilerCurrentStatus = TGCompilerStatus.Initialized;   //still fairly valid
+					return;
```
Cyberboss: One more thing, both the pre and post compile hooks should write to the log e.g.
```
TGServerService.WriteWarning("Precompile hook failed!", TGServerService.EventID.DMCompileError);
```

TGServerService/PreactionHandler.cs
```diff
@@ -0,0 +1,66 @@
...
+			};
+			process.Start();
+			process.WaitForExit();
+
+			var stdout = process.StandardOutput.ReadToEnd();
+			var stderr = process.StandardError.ReadToEnd();
+
+			TGServerService.WriteInfo(
+				String.Format("Preaction Event ran. Stdout:\n{0}\nStderr:\n{1}", stdout, stderr),
+				process.ExitCode == 0 ? TGServerService.EventID.PreactionEvent : TGServerService.EventID.PreactionFail
```
Cyberboss: This should mention the event path/name

TGServerService/ServerService.cs
```diff
@@ -78,6 +78,8 @@ public enum EventID
 			IRCLogModes = 6500,
 			SubmoduleReclone = 6600,
 			Authentication = 6700,
+      PreactionEvent = 6800,
```
Cyberboss: :^)

PR #164, opened 2017-09-11
Fixes the 0 push bug again

MrStonedOne:
This behavior was technically incorrect.

Null indicates a successful commit.

### DIFF
--- a/TGServerService/Repository.cs
+++ b/TGServerService/Repository.cs
@@ -770,7 +770,7 @@ namespace TGServerService
 					var sum = status.Added.Count() + status.Removed.Count() + status.Modified.Count();
 
 					if (sum == 0)   //nothing to commit
-						return null;
+						return "No changes";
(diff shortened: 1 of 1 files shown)


Cyberboss:
You'll have to add this as a separate check to TGControlPanel/ServerPage.cs as well, because it assumes any string returned from these functions is an error and shows a message box. That's also in the definition of PushChangelog() API.

MrStonedOne:
Clearly the issue is we are using return values to signify what exceptions are suppose to be used for.

Cyberboss:
Yeah but sadly we can't throw exceptions through the boundary

MrStonedOne:
Not with that attitude we can't.

MrStonedOne:
By abusing casting, typeless vars (object), and ferry objects, we could easily support throwing exceptions thru the boundary.

PR #181, opened 2017-09-18
Adds a method to set the URL of the cert you want to use

Cyberboss:
Boi


JamieH:
@Cyberboss can we get Letsencrypt? <3 https://github.com/ebekker/ACMESharp

Cyberboss:
Actually, it's all in the windows port bindings and I'm just a fuckwit

PR #192, opened 2017-09-20
Renames ITGServiceBridge to ITGInterop. Can no longer be spoofed by programs outside of DD. Makes authentication saner

Cyberboss:
The interop call is now restricted to the same user the service runs as


Cyberboss:
DBM the rain falls down on the face of the windows identity its actually a groupprince not a user principal

PR #194, opened 2017-09-20
Removes /tg/ specific shims. Adds configuration of static folders and DLLs

### LINKED ISSUE #189: Give repositories a file that can specify config, data dirs, and dlls used
/config, /data and libmysql.dll shouldn't be hard coded

Cyberboss:
Goodbye CP config page, you were always shit.

Added TGS3.json, see example in DMAPI/TGS3.json

This should be able to run any byond game now, provided the .dme is in the root directory and they have a properly configured TGS3.json 

Mostly kept the changelog stuff in code cause it was too pretty.

Closes #189


Cyberboss:
Now that I think about it, the .dme doesn't even have to be in the root if people just add slashes to Project Path:

TGS3.json
```diff
@@ -0,0 +1,58 @@
+{
+  "documentation": [
+    "Place this file in the root of your repo in order to configure TGS3 per repository options",
+    "This file is not required, it merely configures changelog generation, static directories, and .dlls",
+    "All keys except the ones below are ignored by the service"
+  ],
+  "changelog": {
```
JamieH: Use the preaction framework :^)

Cyberboss: You're not wrong, I'm just laz... oh I thought of how to do this!

Cyberboss: Actually no, I'll clean this up after 3.1

PR #196, opened 2017-09-20
Config editing now impersonates the calling windows user. Moves MoveServer() to ITGAdministration

Cyberboss:
As opposed to reading them as LOCAL_SYSTEM and always having access

This allows you to use ACLs to properly restrict who can access which file

Also it was kind of ridiculous not having one of the most destructive server commands behind a gate


Cyberboss:
@MrStonedOne This is probably of interest to you

PR #243, opened 2017-10-12
Ability to publish testmerge commits to remote

Cyberboss:
Just for reference so they can be browsed on github


MrStonedOne:
There is nowhere near enough info on this in this pr.

Is it optional? does it push to master? (I'd assume not, but i can't understand git.net's function/method layout)

Cyberboss:
It's going to be optional yes, the way it works it it pushes a branch to github then immediately deletes it so the commits exist online and can be linked to without 404ing

MrStonedOne:
cool now get @lzimann to exempt the branch name from zbot

Cyberboss:
`___TGS3TempBranch`

Cyberboss:
This isn't working because I can't figure out how to delete a remote branch using LibGit2Sharp

PR #248, opened 2017-10-13
Server instancing

Cyberboss:
Oh boy, here I go instancing again...


Cyberboss:
This is done, just waiting for the version bump

PR #264, opened 2017-10-21
Add missing minibot.py to TGS2

Cyberboss:


Cyberboss:
Wait, no it gets it from the repo

PR #275, opened 2017-10-24
Removes old process priority compiler shennanigans

Cyberboss:
We just use Suspend and Resume now


MrStonedOne:
Actually, could you make it so the dd process is High priority. TGS2 did because it gave a minor edge in performance by forcing windows processes to run on other cores.

Cyberboss:
Will do

PR #278, opened 2017-10-29
BYOND updates trigger a recompile

Cyberboss:
Duh

### DIFF
--- a/TGServerService/ServerInstance/Byond.cs
+++ b/TGServerService/ServerInstance/Byond.cs
@@ -326,6 +326,7 @@ namespace TGServerService
 					Program.DeleteDirectory(ByondDirectory);
 					Directory.Move(StagingDirectoryInner, ByondDirectory);
 					Program.DeleteDirectory(StagingDirectory);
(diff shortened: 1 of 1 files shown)


MrStonedOne:
So, protip, you can freely compile with the new version while waiting for the round to end, just as long as you wait for the compile to complete before setting the flag that states you are waiting for the round to end.

This way you can also avoid any player visible issues if the compile fails

Cyberboss:
This will have to be delayed until I rewrite ServerInstance

MrStonedOne:
Why? we're in the middle of a beta cycle on byond, so this is the exact time something like this should be made and improved and tested, because its the exact time it will actually be used =P

Cyberboss:
Blame my shitty initial architecture

Cyberboss:
In all honesty though, it's about the same amount of work to make a cruddy hack that just works as it is to restructure the BYOND component properly

Cyberboss:
The fix as is wouldn't work either, it'd just recompile using the previously installed version

PR #273, opened 2017-10-24
Reconverts the project to MSIL. Adds seperate bridge .dll

Cyberboss:
Converting to x86 was the lazy way to do it. This is the proper way

Fixes #311

PR #307, opened 2017-11-02
Adds Mono support to the Interface

Cyberboss:
Let's see how this goes...


Cyberboss:
Closing this because once #273 is merged, TGServiceInterface will be compilable by default on mono and the real issue is upstream which doesn't yet support IRequestSessionChannel (needed for windows authentication to work). Gonna dive into mono source to see if I can hack around it.

PR #337, opened 2017-11-10
Adds the test merge manager

Cyberboss:
untitled

Need I say more?


Jordie0608:
I looks like it'd be annoying to snoop out a specific PR, especially if it's not recent; could there be a filter function? Or maybe have the list start empty and allow users to input PR #s with an option to import the list of PRs. It'd be less convenient than current if you want to test a single older PR by having to scroll through the list.

Cyberboss:
The number option still exists on the repo tab

Jordie0608:
Ok, what about for multiple #?

Cyberboss:
The repos merge PR button is different than the one on the (current) server
page in that it doesn't trigger a compile. So you can merge as many as you
want before compiling.

Jordie0608:
I mean for scrolling through the test-merge manager's list of PRs to find multiple old ones.

Cyberboss:
I mean, it'd be nice but then I'd start have to require the user to login
with GitHub/provide an API key so the request rate isn't limited.

MrStonedOne:
I mean, I already have a few api keys laying around. It's not like It would be hard to make a new one for this for our use, and the same applies 

Anyways. This was the system I had in mind:

Ideally, there would be a seperate subform listing all current test merges, with the ability to reorder them,selectively update a section of them.

Adding a test merge should open this subform. (as an aside, people frequently testmerge closed prs, this gui is nice but an option to add a test merge by pr number forcing it to show up on this list would be helpful)


Every time the server is updated, it should do a full resetting update then apply all test merges in order (except any that are now merged on master, that should automatically remove a test merge).

Updating the code shouldn't ever update a test merge's commit, that should be a seperate button.

We could also make an endpoint in the github webhook script that lets you give it a pr number and it returns the changelog. Its a bit more snowflakey since most of tgs3 is designed to be generic, but it would let us have a special section in the changelog for testing changes

Cyberboss:
Currently the GitHub API (https://github.com/OctoKit/OctoKit.Net) is included only in the Control Panel executable. Since that's the only thing that uses it, I'd rather not include it in the service installation. I suppose I can dynamically prompt the user to login if they exceed the anonymous request rate of 60/per hour and then cache their credentials for future requests.

I'll work on implementing the panels @MrStonedOne suggested.

MrStonedOne:
I wonder if there is a way to get a auth token from an api key. we could allow a server side api key with read only permissions, and generate short term sessions sent to clients. It would still require the server have the api, but if its classified to its own dll that wouldn't be a big deal.

Cyberboss:
Okay so it takes a lot more than 60 anonymous request before github rate limits you, and it seems to refresh a lot sooner than 1 hour (every time I launched the debugger I had to spam click) so it's not as big a concern as I thought it'd be. In the rare case that the rate limit is hit I added a simple login prompt, 

untitled

You can provide an API key or your credentials. If you provide your credentials, it creates and saves an encrypted public access API key as you can see in the picture and uses that.

Cyberboss:
Any PR with a label that contains the string "test" (case-insensitive) will be prioritized

untitled

Cyberboss:
The changelog thing is beyond the scope of this PR so you should probably put in a FR for it

Cyberboss:
The update server button has been modified according to @MrStonedOne's description. Entries in the PR list that are merged at a different SHA than the remote repo's PR will have a duplicate entry, giving users the option to either keep a current merge or update it.
untitled

Cyberboss:
Alright I think it's up to snuff

PR #344, opened 2017-11-12
ChatSetupInfo lists are sanity checked

Cyberboss:
There's actually an unavoidable bug in the Control Panel where the textboxes keep adding newlines and feeding back into the chat info. This prevents that


Cyberboss:
CI is borked

PR #362, opened 2017-11-14
Allows the interface timeout to be configured

Cyberboss:


Cyberboss:
30s is fine, the problem is doing shit on the mainline request thread

PR #425, opened 2017-11-28
Gets the interface working on Mono

### LINKED ISSUE #312: Interface chokes somewhere in System.ServiceModel on mono
The TransportWithMessageCredential flag is what we use to allow users to send their windows credentials to the service. However, it seems this throws a not implemented exception in mono:

untitled

Cyberboss:
Fixes #312 

untitled

Let me explain what's going on here.

The problem in the current Interface on mono is that the TransportWithMessageCredential enum, which attaches Windows credentials to a SOAP header, is unimplemented. We can work around this by attaching the username and password to our own custom headers, which are also protected by SSL. Once that's done, we override the ServiceAuthenticationManagers server side to turn the username and passwords into WindowsIdentitys, which are released after the operation completes, and then inject them into the WCF security context.

TL;DR: Mono works, android is next

### DIFF
--- a/TGS.Interface/AuthenticationHeaderApplicator.cs
+++ b/TGS.Interface/AuthenticationHeaderApplicator.cs
@@ -1,4 +1,6 @@
-﻿using System.ServiceModel;
+﻿using System.Collections.Generic;
+using System.Reflection;
+using System.ServiceModel;
 using System.ServiceModel.Channels;
 using System.ServiceModel.Dispatcher;
 using System.ServiceModel.Description;
@@ -16,7 +18,16 @@ namespace TGS.Interface
 
 		public void ApplyClientBehavior(ServiceEndpoint endpoint, ClientRuntime clientRuntime)
 		{
-			clientRuntime.ClientMessageInspectors.Add(this);
+#if __MonoCS__
+			var prop = clientRuntime.GetType()
+									.GetTypeInfo()
+									.GetDeclaredProperty("MessageInspectors");
+			var inspectors = (ICollection<IClientMessageInspector>)prop
+						.GetValue(clientRuntime);
+#else
+			var inspectors = clientRuntime.ClientMessageInspectors;
+#endif
+			inspectors.Add(this);
 		}
 
 		public object BeforeSendRequest(ref Message request, IClientChannel channel)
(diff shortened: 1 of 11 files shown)


TGS.Interface/ServerInterface.cs
```diff
@@ -393,23 +384,31 @@ ChannelFactory<T> CreateChannel<T>(string instanceName)
...
+				var applicator = new AuthenticationHeaderApplicator(LoginInfo);
+				var t = Type.GetType("Mono.Runtime");
+				KeyedCollection<Type, IEndpointBehavior> behaviours;
+				if (t != null)
+				{
+					var prop = res.Endpoint.GetType()
+					 .GetTypeInfo()
+					 .GetDeclaredProperty("Behaviors");
+					behaviours = (KeyedCollection<Type, IEndpointBehavior>)prop
+									 .GetValue(res.Endpoint);
```
Cyberboss: .

TGS.Interface/AuthenticationHeaderApplicator.cs
```diff
@@ -0,0 +1,64 @@
...
+		{
+			var t = Type.GetType("Mono.Runtime");
+			ICollection<IClientMessageInspector> inspectors;
+			if (t != null)
+			{
+				var prop = clientRuntime.GetType()
+									.GetTypeInfo()
+									.GetDeclaredProperty("MessageInspectors");
+				inspectors = (ICollection<IClientMessageInspector>)prop
+							.GetValue(clientRuntime);
```
Cyberboss: .

TGS.Server/Instance/Config.cs
```diff
@@ -15,9 +15,9 @@ sealed partial class Instance : ITGConfig
 		object configLock = new object();	//for atomic reads/writes
 
 		/// <inheritdoc />
-		[OperationBehavior(Impersonation = ImpersonationOption.Required)]
 		public string ReadText(string staticRelativePath, bool repo, out string error, out bool unauthorized)
 		{
+			Server.BeginImpersonation();
```
Cyberboss: We give up attribute based impersonation, but it's more than worth it

Cyberboss:
Travis

Cyberboss:
Inb4 everything breaks

PR #444, opened 2017-12-06
Run garbage collection after a bridge call

Cyberboss:
Fucking it never ran wtf


Cyberboss:
I'm not merging this until I confirm this is the issue

PR #463, opened 2017-12-16
Temp branch name is now derived from Instance name

Cyberboss:
This should fix the race conditions

@lzimann you might want to add `___TGS3TempBranchBagil`, `___TGS3TempBranchSybil`, and `___TGS3TempBranchTerry` to Zbots ignore list


Cyberboss:
Don't like this solution, need to switch to force pushes

PR #465, opened 2017-12-27
Watchdog heartbeat topics

Cyberboss:
Every 10s the watchdog sends a ping to DD. If 10 of these fail, it restarts


MrStonedOne:
Needs a chat command to override or speed up, needs to be configurable and disableable.

PR #504, opened 2018-04-25
Work on the new watchdog

Cyberboss:
I will use less inline lambdas
I will use less inline lambdas
I will use less inline lambdas
I will use less inline lambdas
I will use less inline lambdas
I will use less inline lambdas
I will use less inline lambdas
I will use less inline lambdas
I will use less inline lambdas
I will use less inline lambdas
I will use less inline lambdas
I will use less inline lambdas
I will use less inline lambdas
I will use less inline lambdas


Cyberboss:
This is a lot more than just the watch dog

PR #528, opened 2018-06-27
Save configuration after modification

vuonojenmustaturska:
Would this work? Getting the `get;` autoproperty-generated backing field for modification in a custom `set {}` *IS* possible, but really really shitty.

### DIFF
--- a/TGS.Server/InstanceConfig.cs
+++ b/TGS.Server/InstanceConfig.cs
@@ -19,109 +19,210 @@ namespace TGS.Server
 		/// </summary>
 		ulong Version { get; }
 
-		/// <summary>
-		/// The name of the <see cref="Instance"/>
-		/// </summary>
-		string Name { get; set; }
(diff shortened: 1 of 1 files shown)


vuonojenmustaturska:
why did i think i was modifying a class...

vuonojenmustaturska:
How do you handle config loading? Will this replace the config with default values after reading name?

vuonojenmustaturska:
image

Cyberboss:
Mentioned this in IRC but posting this here for reference. Since the config uses JSON.Net's auto serializer, these changes may affect how those are written out and I'll have to confirm they dont

MrStonedOne:
It shouldn't. the auto serializer only serializes public members and all of those work properly with getters and setters.

Thats from my experience with the soap serializer at least.

Cyberboss:
im_drunk_and_what_could_go_wrong.ogg

PR #533, opened 2018-07-07
Revert "Save configuration after modification"

Cyberboss:
Reverts tgstation/tgstation-server#528

Regrettable a blanket solution like this won't work as it causes weirdness with creating and importing instances trying to reset things back to the default values. I'll just defer to calling Save() everywhere of importance

Fixes #533


Cyberboss:
I.e. https://file.house/okXc.png he wasn't trying to import that folder at all

PR #535, opened 2018-07-09
Update Repo URL in README and GUI

Coolguy3289:
I've added the appropriate notes in the README about setting the python path, as well as the way a Repo URL needs to be inputted. I also changed the default repo URL in the Control Panel GUI to accurately reflect these changes.


Cyberboss:
https not working is an issue but this is whatever so

PR #536, opened 2018-07-09
Changed background color to lighter gray

Coolguy3289:
I've changed the System.Drawing.Color from a dark gray with black text, to a lighter gray making the text easier to read.


Cyberboss:
Could you post a pic please?

Coolguy3289:
It's just a bit brighter
https://imgur.com/a/dDsJeft
However, I realized when I ran the build on my home PC, that the issue I was having is only on the server, the buttons are inheriting the background color and making it hard to see the black text.

PR #544, opened 2018-07-17
Watchdog event handlers

Cyberboss:
@MrStonedOne you might wanna stay subbed to this for feedback

The watchdog monitor has 6 reasons to activate, listed [here](https://github.com/Cyberboss/tgstation-server/blob/WatchdogErrorStates/src/Tgstation.Server.Host/Components/Watchdog/MonitorActivationReason.cs)

It also has a few *modifiers* to it, noted by the `bool`s in [this file](https://github.com/Cyberboss/tgstation-server/blob/WatchdogErrorStates/src/Tgstation.Server.Host/Components/Watchdog/MonitorState.cs)

The watchdog will handle each event as they occur in the sequence listed [here](https://github.com/Cyberboss/tgstation-server/blob/WatchdogErrorStates/src/Tgstation.Server.Host/Components/Watchdog/Watchdog.cs#L330) (Order is debatable)


Cyberboss:
This needs to be tested thouroughly anyway, so we'll cross that bridge when we come to it

PR #547, opened 2018-07-19
Removes DD check from repo recloning

### LINKED ISSUE #546: You should not have to stop DD to reclone the git folder.
@Cyberboss it never ends

Cyberboss:
What could go wrong?

Fixes #546

### DIFF
--- a/TGS.Server/Instance/Repository.cs
+++ b/TGS.Server/Instance/Repository.cs
@@ -390,8 +390,6 @@ namespace TGS.Server
 					if (!CompilerIdleNoLock())
 						return "Compiler is running!";
 				}
-				if (DaemonStatus() != DreamDaemonStatus.Offline)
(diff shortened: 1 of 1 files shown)


TGS.Server/Instance/Repository.cs
```diff
@@ -390,8 +390,6 @@ public string Setup(string RepoURL, string BranchName)
 					if (!CompilerIdleNoLock())
 						return "Compiler is running!";
```
MrStonedOne: Why is this an error? Compiling doesn't require access to the repo once the copy happens.

Not that it will likely ever come up.

Cyberboss: yeah, it's killing a fly with a machine gun but it's the cleanest method without refactoring compiling states into something other than a bool

PR #566, opened 2018-08-15
Removes clients

Cyberboss:
THIS IS NOW AN API


Rockdtben:
Sweet. So we discuss soon.

Cyberboss:
Docker build failed because of a base image change, will sort out later

PR #652, opened 2018-09-15
Overhaul of CompileJob database insertion and consumption

### LINKED ISSUE #651: There's a small window of time where a CompileJob can be retrieved while it's Job hasn't finished

Cyberboss:
Fixes #651 

`CompileJob`s had a weird spot right after they were completed where their `Job`s weren't but several things assumed they were and they existed in the database. This aims to remedy that

`Models.Job` now has an unmapped property called `PostComplete` that will fire after the `Job` has been committed to the db. The results of this call only affect `StoppedAt`, `Cancelled`, and `ExceptionDetails` if it was exceptional and won't run at all if the job before it was exceptional.

The JobManager itself was made responsible for calling the `IDatabaseContext.Save()` that commits `CompileJob`s and then runs a `PostComplete` task which sends it to the `ICompileJobConsumer`.

The `IInstance` auto update and `DreamMakerController` compile processes were merged into one function in `IInstance`

Auto updates now use the `IJobManager` for separate repository and compile steps.

The watchdog no longer tries to guess the admin user's ID


Cyberboss:
This may break things but vOv

Issue #597, opened 2018-09-06
Watchdog finalization


Cyberboss:
TODO

- [ ] Write unit tests that cover each activation state
- [ ] Write integration tests that cover each activation state
- [x] Smoke test it by intentionally rapid fire killing DD processes with Task Manager and fix issues

Cyberboss:
@steamp0rt is helping a ton with this by using it at sunset

PR #708, opened 2018-09-24
NetworkPromptReaper A.K.A. where's your god now DreamDaemon

### LINKED ISSUE #707: No natural way to avoid the port change popup dialog on Windows
I asked Lummox about this, it ALWAYS runs when you call `world.OpenPort()`.

Of course this messes up the entire hotswap procedure.

I'm working on a patch for the mfc120.dll that ships with BYOND. At first we just tried to always positive flag the return value of `CDialog::DoModel` but it seems that prevents the entirety of DreamDaemon from running. So two things need to be done:

~~- [ ] Change the return value patch to a whole separate .dll shim with minimal psuedo as such:~~

```c++
auto ourDll = LoadLibraryA("TGS_Shims.dll");
auto targetProc = GetProcAddress("DoModalShim")
```
```asm
call targetProc
jnz eax, <rest of original function>
mov eax, 6
jmp <function exit>
```

OK, that's retarded, our best bet is probably to recompile mfc120.dll from source and just search in there. But i'm not sure if byond will be able to link it properly reee...

Cyberboss:
User32 hacks to find that shitty prompt and accept it whenever it appears.

Caveat is since it searches active windows globally, it might not work if you have another window open titled "Network Accessibility". So just don't do that, lmao.

Fixes #707 
Connects #597

### DIFF
--- a/src/Tgstation.Server.Host/Components/Watchdog/NetworkPromptReaper.cs
+++ b/src/Tgstation.Server.Host/Components/Watchdog/NetworkPromptReaper.cs
@@ -1,0 +1,191 @@
+﻿using Microsoft.Extensions.Hosting;
+using Microsoft.Extensions.Logging;
+using System;
+using System.Collections.Generic;
+using System.ComponentModel;
+using System.Linq;
+using System.Runtime.InteropServices;
+using System.Text;
+using System.Threading;
+using System.Threading.Tasks;
+using Tgstation.Server.Host.Core;
+
+namespace Tgstation.Server.Host.Components.Watchdog
+{
+	/// <inheritdoc />
+	sealed class NetworkPromptReaper : IHostedService, INetworkPromptReaper, IDisposable
+	{
+		/// <summary>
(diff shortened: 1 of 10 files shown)


src/Tgstation.Server.Host/Components/Watchdog/NetworkPromptReaper.cs
```diff
@@ -0,0 +1,191 @@
...
+							continue;
+						}
+
+						var windowText = stringBuilder.ToString();
+						if (windowText == "Yes")
+						{
+							//smash_button_meme.jpg
+							//yes send it five times for good measure
+							logger.LogTrace("Sending \"Yes\" button clicks...");
+							for (var J = 0; J < 5; ++J)
```
MrStonedOne: Please make the 5 here a define.

Cyberboss:
So it works, revealed more bugs in the process

Cyberboss:
This works so well it's scary

PR #833, opened 2018-12-10
Fix the DMAPI bug

Cyberboss:


Cyberboss:
:cl:
Fixed a runtime in the DMAPI related to version selection.
/:cl:

PR #862, opened 2019-02-19
The watchdog will only restart the server if it exits with a non-zero exit code

MrStonedOne:
:cl:
The watchdog will only restart the server if it exits with a non-zero exit code
/:cl:

[Why]: # (Please add a short description [two lines down] of why you think these changes would benefit the game. If you can't justify it in words, it might not be worth adding.)

So we can make a shutdown server verb.


MrStonedOne:
A graceful dd shutdown has a non-zero exit code, so this won't work.

PR #875, opened 2019-12-29
Adds parsing of port for MySQL/MariaDB databases in SetupWizard

bobbahbrown:
Does as it says on the tin, previously was impossible to set a non-default port.

🆑
Fixed being unable to set the MYSQL/MariaDB port with the setup wizard
/🆑


src/Tgstation.Server.Host/Core/SetupWizard.cs
```diff
@@ -166,8 +167,18 @@ async Task<DatabaseConfiguration> ConfigureDatabase(CancellationToken cancellati
...
+				var serverPort = 3306U;
 				if (String.IsNullOrWhiteSpace(serverAddress))
 					serverAddress = null;
+				else
+				{
+					var m = Regex.Match(serverAddress, @"^(?<server>.+):(?<port>[0-9]+)$");
+					if (m.Success)
+					{
+						serverAddress = m.Groups["server"].Value;
+						serverPort = uint.Parse(m.Groups["port"].Value, CultureInfo.InvariantCulture);
```
Cyberboss: Use `unit.TryParse` here and handle the failure case.

src/Tgstation.Server.Host/Core/SetupWizard.cs
```diff
@@ -166,8 +167,18 @@ async Task<DatabaseConfiguration> ConfigureDatabase(CancellationToken cancellati
 				await console.WriteAsync(null, true, cancellationToken).ConfigureAwait(false);
 				await console.WriteAsync("Enter the server's address and port (blank for local): ", false, cancellationToken).ConfigureAwait(false);
 				var serverAddress = await console.ReadLineAsync(false, cancellationToken).ConfigureAwait(false);
+				var serverPort = 3306U;
```
Cyberboss: Can we make this a `uint?` instead to avoid setting the default (which is done under the hood if it isn't present.

src/Tgstation.Server.Host/Core/SetupWizard.cs
```diff
@@ -239,6 +250,7 @@ void CreateTestConnection(string connectionString)
 					var csb = new MySqlConnectionStringBuilder
 					{
 						Server = serverAddress ?? "127.0.0.1",
+						Port = serverPort,
```
Cyberboss: Then we can conditionally set this on the next line:
```c#
if (serverPort.HasValue)
    csb.Port = serverPort
```

src/Tgstation.Server.Host/Core/SetupWizard.cs
```diff
@@ -166,8 +167,18 @@ async Task<DatabaseConfiguration> ConfigureDatabase(CancellationToken cancellati
 				await console.WriteAsync(null, true, cancellationToken).ConfigureAwait(false);
 				await console.WriteAsync("Enter the server's address and port (blank for local): ", false, cancellationToken).ConfigureAwait(false);
 				var serverAddress = await console.ReadLineAsync(false, cancellationToken).ConfigureAwait(false);
+				var serverPort = 3306U;
 				if (String.IsNullOrWhiteSpace(serverAddress))
 					serverAddress = null;
+				else
```
Cyberboss:
```suggestion
				else if (databaseConfiguration.DatabaseType != DatabaseType.SqlServer)
```
The ConnectionStringBuilder for SQL Server includes its port in the `DataSource` property.

Cyberboss: Or maybe rename `serverPort` to `mySQLServerPort`

bobbahbrown:
Just as a heads up this isn't dead I just have been busy the past two days will fix soon

PR #904, opened 2020-03-23
Add SQLite support

### LINKED ISSUE #885: SQLite support
The issue was with re-entrancy (Which we do a lot of). We should reinvestigate if it has been dealt with upstream.

Cyberboss:
While this would be amazing to have, it currently has issues with the database becoming locked with nested transactions. This is shown with the failing integration tests.

Does the bulk of the work for #885, see that issue for the blocker


tests/Tgstation.Server.Tests/TestingServer.cs
```diff
@@ -79,7 +80,7 @@ public void Dispose()
 		public async Task RunAsync(CancellationToken cancellationToken)
 		{
 			Task runTask = realServer.RunAsync(cancellationToken);
-
+
```
Cyberboss: AHHHHHHHH

tests/Tgstation.Server.Tests/UsersTest.cs
```diff
@@ -0,0 +1,60 @@
...
+			}
+			finally
+			{
+				ThreadPool.SetMinThreads(defaultMinWorker, defaultMinCompletion);
+			}
+
+			Assert.AreEqual(RepeatCount, tasks.Select(task => task.Result.Id).Distinct().Count(), "Did not receive expected number of unique user IDs!");
+		}
+	}
+}
```
Cyberboss: AAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAHHHHHHHHHHHHHHHHHHHHHHHHHHH

Cyberboss:
Merging this because it has good shit (tm) in it. Disabled the sqlite support for now

PR #977, opened 2020-05-12
Fix MySql timeouts

Cyberboss:
:cl:
Fixed a rare NullReferenceException when launching DreamDaemon.
Enabled connection resiliency for MySQL connections. This was causing heavy instability across the server.
Fixed a potential issue where chat bots could report the watchdog was starting instead of restarting.
Fixed deployment messages showing an extra `.0` after the BYOND version.
/:cl:


Cyberboss:
Travis makes no sense right now

PR #1031, opened 2020-05-31
Day zero fixes

Cyberboss:
:cl:
You can now change the watchdog heartbeat interval and startup timeout without triggering a soft restart.
Fixed a watchdog crash loop that could potentially occur during the commit phase of deployment.
Changing the auto-update interval will no longer cancel pull/deploy jobs that are in progress.
Fixed an auto update failure that could occur if auto updates began with the latest origin commit and test merges active.
Fixed a crash that could occur as a result of a race condition when the watchdog reattaches.
Fixed a watchdog crash lop that could potentially occur when updating DreamDaemon settings.
Fixed a Windows watchdog crash that would occur after the first deployment after a reattach.
Fixed the watchdog detach event not being sent to the DMAPI.
/:cl:

:cl: HTTP API
Patch version bump to facilitate implementation changes.
/:cl:

### DIFF
--- a/build/Version.props
+++ b/build/Version.props
@@ -3,8 +3,8 @@
     <!-- This is the authorative version list -->
     <!-- Integration tests will ensure they match across the board -->
     <TgsCoreVersion>4.3.0</TgsCoreVersion>
-    <TgsApiVersion>6.5.0</TgsApiVersion>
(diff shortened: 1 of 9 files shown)


Cyberboss:
TODO: Add Integration test for graceful reboot after settings change

PR #1093, opened 2020-08-01
Infinitely cache JS files

Cyberboss:
Requested by @alexkar598


Cyberboss:
@alexkar598 we'll deal with this when we first merge in in the vnext cp

PR #1100, opened 2020-08-10
improves grammar in README

Fikou:
It doesn't look good.


Cyberboss:
Removed your changelog as it's only for end user notifications

PR #1116, opened 2020-09-06
Fix discordconnectionstringbuilder crash

MrStonedOne:
crash if basedMeme was not provided but displaytype was.
:cl: HTTP API
Fix off by one bug when processing discord connection strings with exactly 2 arguments given leading to 500 responses.
/:cl:


bobbahbrown:
naming conventions that make you say hmm

Cyberboss (approved):

PR #1129, opened 2020-10-05
Fixes set-env depreciation 

ZeWaka:

[GitHub bad.](https://github.blog/changelog/2020-10-01-github-actions-deprecating-set-env-and-add-path-commands/) Depreciation bad. CI breaking in the future bad. 

This fix. Gooncoder happy.


Cyberboss:
Meh, GitHub code scanning should have been pushed to master anyway.

Cyberboss (approved):
Looks good

PR #1171, opened 2020-12-16
V3: Adds build directory as parameter to PreCompile.sh

Cyberboss:
@MrStonedOne Turns out I can build an installer for this if needed. How do you want to do it?

Needed for https://github.com/tgstation/tgstation/pull/55373


MrStonedOne:
Installer would be easiest.

PR #1180, opened 2020-12-31
Late hour fixes

Cyberboss:
No changelog, all unreleased fixes.


actioninja:
>aprilfixups
sometimes it be like that

PR #1181, opened 2020-12-31
Updates TGS to webpanel 1.2.0 on the master branch.  The last version to support api 7.x.x

alexkar598:
:cl: Web Control Panel
Updated the webpanel to version 1.2.0. Bringing a few misc improvements
/:cl:

Contains a few changes. nothing big but since support for 7.x.x is going to be dropped. might as well release everything there is in the webpanel now


Cyberboss:
Too close to 4.7 for this to be worth it, sorry

PR #1192, opened 2021-01-09
[READY] Adds an event script for triggering events when DD is ended with TgsEndProcess()

AffectedArc07:
:cl: Core
Added a new event (`WorldEndProcess`) which triggers on TgsEndProcess().
/ :cl:

Useful for automations between rounds (IE: cloning specific log files from that round exactly to a webserver)


src/Tgstation.Server.Host/Components/Events/EventType.cs
```diff
@@ -111,6 +111,12 @@ public enum EventType
...
 		/// </summary>
 		[EventScript("WatchdogLaunch")]
-		WatchdogLaunch
+		WatchdogLaunch,
+
+		/// <summary>
+		/// Inbetween DD restarts if the process has been force-ended by the DMAPI
+		/// </summary>
+		[EventScript("DDEndProcess")]
+		DDEndProcess
```
Cyberboss, line 120: Can you rename this `WorldEndProcess` please?

Cyberboss (changes requested):
So there's two issues with this currently.

1. You call the task and don't await it, which is bad mojo as it leaves any async `Exception`s it may throw lying around un-logged in memory.

You could just `await` it in this context but...

2. You're calling the event from the bridge request context, which means the final HTTP request DD makes to TGS before getting killed. It's also AFTER the call to `process.Terminate()`.

That presents a big issue the `cancellationToken` is likely already triggered (due to killing the process closing the socket and thus the request) so the script will be aborted right as it starts. Even if you move the call before the terminate, you'd be holding the request open for the duration of the script.

To fix all this, I'd move the call to `.HandleEvent` into the `BasicWatchdog` code that handles the server exiting. Namely under https://github.com/tgstation/tgstation-server/blob/fd69326e5823877704fd8d9792ebd3ae2e5efb85/src/Tgstation.Server.Host/Components/Watchdog/BasicWatchdog.cs#L93

Would look like:
```c#
case MonitorActivationReason.ActiveServerCrashed:
	if (Server.TerminationWasRequested)
		await EventHandler.HandleEvent(...).ConfigureAwait(false);

	var exitWord = Server.TerminationWasRequested ? "exited" : "crashed";
```

`WatchdogBase` which `BasicWatchdog` inherits already has an `IEventHandler`, it just needs to be made into a `protected` get property instead of a `readonly` field.

This way, you'll get your desired behavior of:
- DD asks TGS for termination.
- Bridge request kills DD.
- Watchdog wakes up, sees the termination request, and calls the script (if it exists).
- Watchdog will wait until the script completes.
- Watchdog will launch DD again (unless a graceful termination was requested).

And then could you revert the changes to `InstanceFactory`, `SessionControllerFactory`, and `SessionController` please.

src/Tgstation.Server.Host/Components/Events/EventType.cs
Cyberboss: And add a trailing `,`

src/Tgstation.Server.Host/Components/Events/EventType.cs
```diff
@@ -114,9 +114,9 @@ public enum EventType
 		WatchdogLaunch,
 
 		/// <summary>
-		/// Inbetween DD restarts if the process has been force-ended by the DMAPI
+		/// Inbetween DD restarts if the process has been force-ended by the DMAPI (TgsEndProcess())
```
Cyberboss, line 117:
```suggestion
		/// In between DD restarts if the process has been force-ended by the DMAPI (TgsEndProcess()).
```

src/Tgstation.Server.Host/Components/Watchdog/WatchdogBase.cs
```diff
@@ -121,7 +121,7 @@ public WatchdogStatus Status
 		/// <summary>
 		/// The <see cref="IEventConsumer"/> that is not the <see cref="WatchdogBase"/>
 		/// </summary>
-		readonly IEventConsumer eventConsumer;
+		public IEventConsumer eventConsumer;
```
Cyberboss, line 124:
```suggestion
		protected IEventConsumer EventConsumer { get; }
```
And move it up to under line 84

Cyberboss (changes requested):

Cyberboss (changes requested):
You also have to make a DMAPI definition for the event in `src/DMAPI/tgs.dm`.

src/Tgstation.Server.Host/Properties/PublishProfiles/FolderProfile.pubxml
```diff
@@ -0,0 +1,22 @@
+<?xml version="1.0" encoding="utf-8"?>
```
Cyberboss, line 1: Remove please, and maybe add .pubxml to the .gitigore

Cyberboss (changes requested):

src/Tgstation.Server.Host/Properties/PublishProfiles/FolderProfile.pubxml
AffectedArc07: I never made a publish so no idea how this happened

Cyberboss (approved):

tgstation-server:
Your Fix/Feature pull request delta is currently below zero (-2). Maintainers may close future Feature/Tweak/Balance PRs. Fixing issues or helping to improve the codebase will raise this score.

PR #1212, opened 2021-01-28
In which Cyberboss will rapidly lose his remaining sanity while doing battle with GitHub Actions

### LINKED ISSUE #1206: Spurious Integration Test Errors
We get a lot of failed tests that shouldn't because of what I can only imagine to be a timing issue in their writing. It's not reproducible locally (With my specs at least: 3.78GHz, 16GB RAM).

The problem is in the `Tgstation.Server.Tests.Instance.WatchdogTest` test class. That class does a lot of chat/topic/interop/process fuckery to thoroughly exercise the watchdog.

See https://github.com/tgstation/tgstation-server/blob/dev/.github/CONTRIBUTING.md#development-environment for how to setup the integration tests locally.

I'm putting a standing 50$ CAD bounty for whoever is able to fix this issue _without_ losing test functionality.

Cyberboss:
Pray to god we will fix #1206

Expecting this first commit to fail to get a sense of the cause


Cyberboss:
Updates in issue

PR #1217, opened 2021-01-31
Update ControlPanelVersion.props so it can support node 15

alexkar598:
No CL because this only changes compile stuff, lmk if i need one anyways


Cyberboss:
:cl: DMAPI
Fixed a compile error in the V3 API.
/:cl:

PR #1283, opened 2021-06-15
Fix single server swarms update deadlock

Cyberboss:
Credit @ZeWaka for finding this bug

:cl:
Fixed a deadlock when trying to update a server swarm that only had a controller and no nodes.
/:cl:


ZeWaka:
I assume a manual update will be required for the affected versions?

Cyberboss:
> 
> 
> I assume a manual update will be required for the affected versions?

Yes, or you can disable swarm mode in the config, restart the server, update, and then re-enable swarm mode in the config and restart. Note this isn't a fix for your `Could not load assembly` issue.

PR #1348, opened 2021-11-05
Security check

TheFakeElon:
[Base Branch]: master

[Release Notes]: fixes a security flaw in tgs3
tgstation/tgstation#62568
:cl:
fixes a security flaw
/:cl:

### DIFF
--- a/src/DMAPI/tgs/v3210/api.dm
+++ b/src/DMAPI/tgs/v3210/api.dm
@@ -54,6 +54,8 @@
 	return ""
 
 /datum/tgs_api/v3210/proc/file2list(filename)
+	if(IsAdminAdvancedProcCall())
+		CRASH("Attempted to read file via admin call")
 	return splittext(trim_left(trim_right(file2text(filename))), "\n")
 
 /datum/tgs_api/v3210/OnWorldNew(minimum_required_security_level)


ExcessiveUseOfCobblestone (approved):

LemonInTheDark:
Mothblocks is blind #1349

LemonInTheDark:
wait wrong version

MCterra10:
can we merge this already?

LemonInTheDark:
no, because it breaks servers that don't have that proc (Insert better reason cyber will give here)
This whole thing is so mindnummingly dumb I wish to return to an era before security posting

src/DMAPI/tgs/v3210/api.dm
```diff
@@ -54,6 +54,8 @@
 	return ""
 
 /datum/tgs_api/v3210/proc/file2list(filename)
+	if(IsAdminAdvancedProcCall())
+		CRASH("Attempted to read file via admin call")
```
alexkar598, line 58: This proc is not part of DMAPI

alexkar598 (changes requested):

AffectedArc07:
The real fix here is to stop using deprecated software which was coded in one afternoon while cyberboss was stoned
image

AffectedArc07:
Putting another post. You can mitigate this on your own codebases outside of the DMAPI. Here are a few ways in no particular order:
- `GENERAL_PROTECT_DATUM(/datum/tgs_api/v3210)`
- Defining a proc after this to make inheritence work. Just put this in your codebase after your DMAPI folder
```
/datum/tgs_api/v3210/file2list(filename)
	if(IsAdminAdvancedProcCall())
		CRASH("Attempted to read file via admin call")
	..()
```

Cyberboss:
Various reasons/alternative already stated, plus you could always redefine file2text. Modifying external library code is not the wei here.

Cyberboss:
Revisionist history 
image

Issue #1301, opened 2021-08-20
Additional input for DreamDaemon command line options


Wirewraith:
**Is your feature request related to a problem? Please describe.**
There is currently no intended way to add arbitrary command line options to the DreamDaemon process. (e.g. `-logself`).

**Describe the solution you'd like**
Just a new input that takes a string that will be appended to the command.

**Describe alternatives you've considered**
It's possible to use the "additional parameters" input to achieve this functionality, however that field is really only intended for adding to the `-params` option, not for adding arbitrary other options.

Issue #1107, opened 2020-08-19
Pull requests from forks fail due to not having some secrets set


Cyberboss:
It can't live without these env vars which are encoded as secrets:

https://github.com/tgstation/tgstation-server/blob/0ae8ad2dfbc0ddbe307e154c4b1b67643f2e06b3/.github/workflows/ci-suite.yml#L14-L17

Cyberboss:
Workaround: create a mirror of PR'd work on the repo to run checks and merge that

Cyberboss:
Example #1129

Cyberboss:
GitHub has a fix for this, something to do with the `pull_request_base` workflow trigger

Cyberboss:
$20 CAD on this

PR #1379, opened 2022-08-19
Add -verbose argument to Dream Daemon by default

out-of-phaze:
:cl:
Adds -verbose argument to Dream Daemon by default.
/:cl:

TGS3 automatically provided the -verbose argument to DD; TGS4 doesn't, limiting the number of runtimes that can be viewed. The -params option only lets you pass params to world.params or whatever, which means adding -verbose there does nothing.

This follows the same logic as #1301, where instead of adding a custom params arg, -logself was added to the default. Also, feature parity with v3 is good, generally.


AffectedArc07:
@Cyberboss Can we push this into `master` as a "Fix missing functionality"?

Cyberboss:
Nice try

Cyberboss (approved):

Cyberboss:
#1107. I ran the tests manually, they passed

PR #1382, opened 2022-08-24
Fix DD response model

AffectedArc07:

:cl: Core
Fixed the DD response model not containing whether the profiler would autostart
/:cl:


Cyberboss (approved):
Technically we should also add a field called `CurrentProfiler` or something to indicate if the current session was launched with the parameter but w/e.

PR #1388, opened 2022-09-25
Set Cache-Control: no-cache for API responses

Cyberboss:
:cl:
`Cache-Control` has been set to `no-cache` for all API responses.
/:cl:


LemonInTheDark (approved):
Does this even work

PR #1422, opened 2022-12-29
Example SystemD service file

### LINKED ISSUE #1420: Linux install needs to auto add a systemD unit and be preferred over docker
**Is your feature request related to a problem? Please describe.**
Yes. The docker install doesnt like hot process reloads, and makes everything more of a pain in the ass than is reasonably required. TGS runs better baremetal, dont kid yourself.

**Describe the solution you'd like**
- Remove docker as the recommendation and add it as an unsupported alternative
- Make the TGS installed automatically install a systemD unit for TGS (theres probably a lib for this, or just manually make a file in `/etc/systemd/system/`)

**Describe alternatives you've considered**
Include a `.service` template in the distribution files that people can modify to their needs

**Additional context**
image

ZephyrTFA:
Closes #1420

### DIFF
--- a/README.md
+++ b/README.md
@@ -37,7 +37,8 @@ If using the console version, run ./tgs.bat in the root of the installation dire
 
 #### Linux (Native)
 
-It is recommended to run TGS using a systemd script. An example script is provided in the base directory, and assumes that TGS is installed under /opt/tgs
+An example systemd script is provided in the base directory, which assumes that TGS is installed under `/opt/tgs`
+If you are experiencing issues using systemd to host TGS and have confirmed all requisites are installed, consider using the docker setup seen below.
 
 The following dependencies are required to run tgstation-server on Linux alongside the .NET Core runtime
 
--- a/src/Tgstation.Server.Host.Console/tgs.service
+++ b/src/Tgstation.Server.Host.Console/tgs.service
@@ -1,5 +1,5 @@
 [Unit]
-Description=TGS Server
+Description=tgstation-server
 After=network.target
 
 [Service]
--- a/src/Tgstation.Server.Host.Console/tgs.service
+++ b/src/Tgstation.Server.Host.Console/tgs.service
@@ -1,6 +1,7 @@
 [Unit]
 Description=tgstation-server
 After=network.target
+After=mariadb.target
 
 [Service]
 ExecStart=bash /opt/tgs/tgs.sh
--- a/src/Tgstation.Server.Host.Console/tgs.service
+++ b/src/Tgstation.Server.Host.Console/tgs.service
@@ -4,7 +4,7 @@ After=network.target
 After=mariadb.target
 
 [Service]
-ExecStart=bash /opt/tgs/tgs.sh
+ExecStart=/bin/bash /opt/tgs/tgs.sh
 Restart=Always
 
 [Install]
--- a/src/Tgstation.Server.Host.Console/tgs.service
+++ b/src/Tgstation.Server.Host.Console/tgs.service
@@ -1,0 +1,10 @@
+[Unit]
+Description=TGS Server
+After=network.target
+
+[Service]
+ExecStart=bash /opt/tgs/tgs.sh
+Restart=Always
+
+[Install]
+WantedBy=multi-user.target


src/Tgstation.Server.Host.Console/tgs.service
```diff
@@ -0,0 +1,10 @@
+[Unit]
+Description=TGS Server
```
Cyberboss, line 2:
```suggestion
Description=tgstation-server
```

README.md
```diff
@@ -37,7 +37,7 @@ If using the console version, run ./tgs.bat in the root of the installation dire
 
 #### Linux (Native)
 
-We recommend using Docker for Linux installations, see below. The content of this parent section may be skipped if you choose to do so.
+It is recommended to run TGS using a systemd script. An example script is provided in the base directory, and assumes that TGS is installed under /opt/tgs
```
Cyberboss, line 40: I still can't recommend this because it can cause unforseen issues that are already accounted for in the Docker version

Cyberboss (changes requested):

README.md
ZephyrTFA:
```suggestion
An example systemd script is provided in the base directory, which assumes that TGS is installed under `/opt/tgs`
If you are experiencing issues using systemd to host TGS and have confirmed all requisites are installed, consider using the docker setup seen below.
```

README.md
```diff
@@ -37,7 +37,8 @@ If using the console version, run ./tgs.bat in the root of the installation dire
 
 #### Linux (Native)
 
-We recommend using Docker for Linux installations, see below. The content of this parent section may be skipped if you choose to do so.
+An example systemd script is provided in the base directory, which assumes that TGS is installed under `/opt/tgs`
+If you are experiencing issues using systemd to host TGS and have confirmed all requisites are installed, consider using the docker setup seen below.
```
ZephyrTFA, line 41: @Cyberboss 
This better?

src/Tgstation.Server.Host.Console/tgs.service
```diff
@@ -0,0 +1,10 @@
+[Unit]
+Description=tgstation-server
+After=network.target
```
AffectedArc07, line 3:
```suggestion
After=network.target
After=mariadb.target
```
99% of linux installs will be using MariaDB as the database backing, and you will want TGS to launch after that.

README.md
```diff
@@ -53,7 +54,7 @@ Note that tgstation-server has only ever been tested on Linux via it's [docker e
 
 #### Docker (Linux)
 
-tgstation-server supports running in a docker container and is the recommended deployment method for Linux systems. The official image repository is located at https://hub.docker.com/r/tgstation/server. It can also be built locally by running `docker build . -f build/Dockerfile -t <your tag name>` in the repository root.
+tgstation-server supports running in a docker container although it is only recommended if you are having issues running it natively. The official image repository is located at https://hub.docker.com/r/tgstation/server. It can also be built locally by running `docker build . -f build/Dockerfile -t <your tag name>` in the repository root.
```
Crossedfall, line 57: I'd still argue that it is the recommended method for running TGS, to be honest.

src/Tgstation.Server.Host.Console/tgs.service
```diff
@@ -0,0 +1,11 @@
...
+After=mariadb.target
+
+[Service]
+ExecStart=bash /opt/tgs/tgs.sh
```
Crossedfall, line 7:
```suggestion
ExecStart=/bin/bash /opt/tgs/tgs.sh
```

Always use explicit paths.

src/Tgstation.Server.Host.Console/tgs.service
```diff
@@ -0,0 +1,11 @@
+[Unit]
+Description=tgstation-server
+After=network.target
+After=mariadb.target
+
+[Service]
```
Crossedfall, line 6: Should have a [type](https://www.freedesktop.org/software/systemd/man/systemd.service.html#Type=) definition.

Crossedfall: Also, is the recommendation really to run this as root? That seems terrible from a best practices perspective.

AffectedArc07: Pretty sure it's needed for process priority elevation stuff, plus some of the event scripts install stuff.

Also the windows service runs as LOCAL SERVICE since you need admin rights to even make symlinks on windows.

ZephyrTFA: > Should have a [type](https://www.freedesktop.org/software/systemd/man/systemd.service.html#Type=) definition.

simple is the default type, and the type we want to run it under since the initial script uses exec instead of spawn

Crossedfall: > > Should have a [type](https://www.freedesktop.org/software/systemd/man/systemd.service.html#Type=) definition.
> 
> simple is the default type, and the type we want to run it under since the initial script uses exec instead of spawn

You should always be explicit.

Crossedfall: > Also the windows service runs as LOCAL SERVICE since you need admin rights to even make symlinks on windows.

LOCAL SERVICE is, as you might gleam from the name, a service account and maintains minimum privileges. It's not a root-adjacent account. 

> Pretty sure it's needed for process priority elevation stuff

That should be handled in a limits.conf file. Alternatively, the user could be given privileges to edit the niceness values of their processes.

> plus some of the event scripts install stuff.

That should not be the default if you're running bare metal. Best practice should be to install everything you need before running TGS as a system service.

Crossedfall: > 
> Correction, its `LOCAL SYSTEM`
> 
> MSFT docs: `An account, used by the service control manager, that has extensive privileges on the local computer and acts as the computer on the network.`

LOCAL SYSTEM **is** a root-adjacent role. Although, it's still restricted in that it only has access to local and domain resources. It can't access external resources, which is not a restriction root has on Linux.

alexkar598: Reminder that 
`AmbientCapabilities=CAP_SYS_NICE`

AFAIK, that covers all the things ***TGS*** needs from running as root.

Note that sample EventScripts for tgstation are built with the expectation of running as root

PR #1450, opened 2023-04-09
Discord replies no longer mention the user.

MrStonedOne:
I don't want this clogging up my mentions inbox (which i check for things i forgot to do) every time i do !tgs. and it will once there are 7 bots on tgs 5 and all 7 of them mention me every time I use a command.

There is no need to mention or ping the user for a response to a message they sent less than a second ago. 

The goal of a mention is to alert the user there is a message they might want to see, but in this case, they kinda know, given they just sent the command.

I likey broke something given im using c# concepts i know nothing about 10 years after the last c# code i wrote.


tralezab:
I've been pestering mso so much his next pr is that Armhulen replies no longer mention the user

MrStonedOne:
@Cyberboss i made it compile, you can handle making it follow lints.

Cyberboss:
Note to self: check this doesn't disable text mentions

Cyberboss:
:cl:
Chat command replying no longer pings the user.
/:cl:

PR #1464, opened 2023-04-29
Add -map-threads parameter support + 510 backwards compatibility and testing

### LINKED ISSUE #1458: Add support for upcoming -map-threads parameter
DreamDaemon launch parameter when not validating.

Cyberboss:
:cl:
Windows firewall rule names for DreamDaemon now include the BYOND version number.
Fix priority topic sends not being able to be interrupted.
/:cl:

:cl: HTTP API
Added the `mapThreads` field to the DreamDaemon model. This can be set to 0 to use the default value or a higher number to be added to DreamDaemon command lines. Only applies to supported BYOND versions.
Added DreamDaemon right 524288 for changing the `mapThreads` field.
/:cl:

Also preliminary DMAPI changes for BYOND backwards compatibility, but running into issues I need to narrow down.

Merge with `[APIDeploy][NugetDeploy]`

Closes #1458


src/Tgstation.Server.Host/Components/Byond/ByondInstallation.cs
```diff
@@ -18,6 +18,9 @@ sealed class ByondInstallation : IByondInstallation
 		/// <inheritdoc />
 		public bool SupportsCli { get; }
 
+		/// <inheritdoc />
+		public bool SupportsMapThreads => Version.Major >= 515 && Version.Minor >= 1605;
```
Cyberboss, line 22: The magic number that may or may not go up

tests/Tgstation.Server.Tests/Live/Instance/ByondTest.cs
```diff
@@ -24,7 +24,7 @@ namespace Tgstation.Server.Tests.Live.Instance
 {
 	sealed class ByondTest : JobsRequiredTest
 	{
-		public static readonly Version TestVersion = new(514, 1588);
+		public static readonly Version TestVersion = new(515, 1605);
```
Cyberboss, line 27: Changing version under test to a newer one with this too

Cyberboss (commented):
Tests will fail until map threads is released

src/Tgstation.Server.Host/Components/Byond/ByondInstallerBase.cs
```diff
@@ -18,6 +18,11 @@ abstract class ByondInstallerBase : IByondInstaller
 		/// </summary>
 		const string CacheDirectoryName = "cache";
 
+		/// <summary>
+		/// The first <see cref="Version"/> of BYOND that supports the '-map-threads' parameter on DreamDaemon.
+		/// </summary>
+		public static Version MapThreadsVersion => new (515, 1605);
```
Cyberboss, line 24: The magic number. I noticed Lummox put out 1605 bins already and they don't have -map-threads so this is subject to change.

MrStonedOne: and again

Cyberboss: Lummox said 1609

Cyberboss:
Need to bisect the version where the BYOND linux bug is fixed

PR #1484, opened 2023-05-04
Test the Basic watchdog and Debug mode on Linux

Cyberboss:
Quadruples the amount of Linux Live tests but that's fine since they mostly work now


Cyberboss:
TODO: Test SqlServer on Linux and the other DBs on Windows

PR #1494, opened 2023-05-21
Nightlies

Cyberboss:
Add midnight UTC scheduled CI. Has to be on the default branch, currently dev.


.github/workflows/ci-suite.yml
```diff
@@ -1,6 +1,8 @@
 name: 'CI'
 
 on:
+  schedule:
+    - cron: 0 0 * * *
```
AffectedArc07, line 5: Schedule this for 23:00

CRON runs at 00:00 are so backlogged ours end up completing at 01:30.

PR #1507, opened 2023-06-02
Fix server applying the deployment low priority config to live servers and the live server priority config to deployments.

MrStonedOne:
This was causing prod live byond servers to run in low priority if `Session:LowPriorityDeploymentProcesses` was enabled.

:cl: Core
Fix server applying the deployment low priority config to live servers and the live server priority config to deployments.
/:cl:

### DIFF
--- a/src/Tgstation.Server.Host/Components/Session/SessionControllerFactory.cs
+++ b/src/Tgstation.Server.Host/Components/Session/SessionControllerFactory.cs
@@ -525,7 +525,7 @@ namespace Tgstation.Server.Host.Components.Session
 
 			try
 			{
-				if (apiValidate)
+				if (!apiValidate)
 				{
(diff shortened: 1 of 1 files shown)


ZephyrTFA:
How the fuck is adding a `!` causing code coverage to change.
Shits broken

MrStonedOne:
Only once case was being checked, and now that leads down a path with more lines?

Cyberboss:
@ZephyrTFA the integration tests (which is the built of coverage) aren't fully deterministic since they (like the MC) rely on task execution order, also the GitHub API, and just general DreamDaemon flakiness

Cyberboss (approved):

Cyberboss:
Parroting, images are just from the codecov report
image

PR #1519, opened 2023-06-06
CI: skip Doxygen builds on forks

ilia-shipitsin:

:cl:
Skip Doxygen build on forks
/:cl:

:cl: Categories
CI/:cl:

[Why]:  small improvement

### DIFF
--- a/.github/workflows/ci-suite.yml
+++ b/.github/workflows/ci-suite.yml
@@ -163,7 +163,7 @@ jobs:
     name: Build Doxygen Site
     runs-on: ubuntu-latest
     needs: start-ci-run-gate
-    if: "!(cancelled() || failure()) && needs.start-ci-run-gate.result == 'success'"
+    if: "!(cancelled() || failure()) && needs.start-ci-run-gate.result == 'success' && github.repository_owner == 'tgstation'"
     steps:
     - name: Checkout (Branch)
       uses: actions/checkout@v3


Cyberboss (changes requested):
We do want to always build the doxygen documents as part of the CI step to check for errors. Could you change this to only prevent the `gh-pages push` step? (Line 189)

Also, instead of hard coding the owner name, let's instead create a new secret that will trigger the push if set? Currently it piggybacks off of `secrets.DEV_PUSH_TOKEN` which is used for releases.

Last thing to note, just FYI. Since this PR is from a fork, you won't see your changes in the PR's actions due to how we use `pull_request_target`. That's not really relevant in the case of this PR, though, so don't worry about it.

Cyberboss (changes requested):
One more thing, since this is a CI improvement. Could you target our stable branch `master` instead of `dev` please?

ilia-shipitsin:
> 

sure

ilia-shipitsin:
well, I'm not sure how to make rule based on `DEV_PUSH_TOKEN`
if we add that to `env` variables, it will lead us to failure on fork if no such secret defined.

(as that workflow already fails on forks, it does not look like a benefit)

Cyberboss:
You can set the secret to an `env` var and then check it with an `if: ${{
env.SECRET_NAME }}` in the step

ilia-shipitsin:
abandoning in favour of https://github.com/tgstation/tgstation-server/pull/1522

PR #1524, opened 2023-06-07
CI: skip code scanning on forks

chipitsine:
:cl:
CI: skip code scanning on forks
/:cl:

:cl: Categories
CI
/:cl:

[Why]: small improvement

### DIFF
--- a/.github/workflows/ci-suite.yml
+++ b/.github/workflows/ci-suite.yml
@@ -58,7 +58,7 @@ jobs:
   analyze:
     name: Code Scanning
     needs: start-ci-run-gate
-    if: "!(cancelled() || failure()) && needs.start-ci-run-gate.result == 'success'"
+    if: "!(cancelled() || failure()) && needs.start-ci-run-gate.result == 'success'" && env.TGS_RELEASE_NOTES_TOKEN != ''
     runs-on: ubuntu-latest
     steps:
     - name: Install Node 12.X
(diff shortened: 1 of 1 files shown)


Cyberboss:
I don't understand why this is necessary. Code Scanning uses the default `GITHUB_TOKEN` which should be present on any repository.

chipitsine:
image


image


I guess it want me to enable some code scanning storage which is not enabled by default

Cyberboss:
It could potentially be a problem related to running in a private fork, which would cause all sorts of problems as it's not really supported. What's the full error you're getting.

chipitsine:
it is public fork

https://github.com/ilia-shipitsin/tgstation-server/actions/runs/5193367388/jobs/9363821773

Cyberboss:
Ahh, I know. The GitHub token must be too restrictive. It needs write access to generate the CodeQL status . You can set it under `settings/actions` (Write used to be the default but they must have changed it).
image

chipitsine:
well, that is cosmetic, indeed.

I can either enable token permission or ignore red status.
I think people do not use Code scanning in forks (but I may be wrong)

Cyberboss:
I've created the above PR to address this as it requires a new repository variable.

chipitsine:
abandoning in favour of https://github.com/tgstation/tgstation-server/pull/1530

PR #1566, opened 2023-06-21
Automatic UAC evaluation for Tgstation.Server.Host.exe

Peliex:
Creates a separate resource manifest for the Tgstation.Server.Host.exe so that the process will automatically elevate to the user's highest permission when ran.  Shows a warning to the user and triggers UAC.

:cl:
On Windows, TGS will request administrator privileges when needed.
/:cl:

Automatic elevation of executable file will prompt UAC even when running on release, this prevents exceptions and errors from occurring because of an improperly configured manifest.  Errors will persist if user fails to provide sufficient permissions.


Cyberboss (changes requested):
Can you change your changelog to just

```
On Windows, TGS will now request UAC elevation if necessary
```

Also remove the second entry, please. It's not needed and not in the correct format

Cyberboss (approved):

PR #1591, opened 2023-07-11
Updates docker image to debian 12

alexkar598:
:cl:
Docker image is now based off of debian bookworm(12).
/:cl:

It's quite silly and funny. Dunno if this is a breaking change or not. TGS itself should keep working but event scripts might feel a bit sad. This is mostly because I want the latest toys from 12


Cyberboss (changes requested):
I'm hesitant to release this change without first running the test suite on whatever base image docker uses. For that, we'll need a separate `Dockerfile`. Running it with just Release, Sqlite, System Watchdog should be sufficient.

Cyberboss:
Fuck it

PR #1594, opened 2023-07-14
Add testing Dockerfile and add to CI

Cyberboss:


Cyberboss:
Spurious errors must be cleaned up first

Cyberboss:
It's not worth it.

PR #1646, opened 2023-09-01
Fixes ambiguous inheritdocs

distributivgesetz:
[Base Branch]: master

[Release Notes]: 

Nothing user-facing

[Why]: make it so my IDE yells at me less loudly

### DIFF
--- a/src/Tgstation.Server.Host/IO/Console.cs
+++ b/src/Tgstation.Server.Host/IO/Console.cs
@@ -7,7 +7,10 @@ using Tgstation.Server.Host.System;
 
 namespace Tgstation.Server.Host.IO
 {
-	/// <inheritdoc />
+	/// <summary>
+	/// Implementation of <see cref="Tgstation.Server.Host.IO.IConsole"/>,
+	/// uses <see cref="global::System.Console"/>.
+	/// </summary>
 	sealed class Console : IConsole, IDisposable
 	{
 		/// <inheritdoc />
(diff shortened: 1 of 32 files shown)


Cyberboss (changes requested):
Changelog is only for user-facing changes so could you remove it please?

Cyberboss (changes requested):
Also can you scope these appropriately so they don't trigger `IDE0001` on VS please?
image

distributivgesetz:
this is gonna take some adjusting lol

src/Tgstation.Server.Host/IO/Console.cs
```diff
@@ -7,7 +7,10 @@
 
 namespace Tgstation.Server.Host.IO
 {
-	/// <inheritdoc />
+	/// <summary>
+	/// Implementation of <see cref="Tgstation.Server.Host.IO.IConsole"/>,
```
Cyberboss, line 11: Missed this one

Cyberboss (changes requested):

src/Tgstation.Server.Host/IO/Console.cs
distributivgesetz: I'm so close to pulling the trigger on Rider I swear to god

Cyberboss (approved):

Cyberboss:
Docker build is broken because of the node version we were using getting too old. #1647 fixes it. Can you merge master once that goes through?

Cyberboss (approved):

Issue #1638, opened 2023-08-23
CI Completion Gate is not working as Intended


Cyberboss:
I think it counts as passed when skipped??

PR #1651, opened 2023-09-13
New Icon integration

Cyberboss:
Remove binary images. Make them generate from .svg on build


Cyberboss:
This should not have been merged, see #1638

PR #1661, opened 2023-10-07
Update to .NET 8 RC1 and Update EFCore

Cyberboss:


Cyberboss:
Need to update CI dotnet

PR #1677, opened 2023-10-18
v5.16.3: Fix PosixWatchdog

### LINKED ISSUE #1680: PosixWatchdog doesn't work
DD/Linux always resolves the symlinks. Needs to be done in a different way.

Current options:

- Hard links
- bindfs

Cyberboss:
:cl:
Fixed issues with DreamDaemon referencing the original true path of instances when using the advanced watchdog on Linux. Hard linked copies of deployment directory structure are now used instead of symlinks.
If a deployment completes on Linux right before a world with a valid DMAPI reboots it may not have enough time to be applied and delayed until the next reboot.
/:cl:

Fixes #1680 

- [x] Rename `ISymlinkFactory` to `IFileSystemLinkFactory`
- [x] Rename `SwappableDmbProvider` to `SymlinkDmbProvider`.
- [x] Extract `WindowsWatchdog` code into abstract `AdvancedWatchdog`.

### DIFF
--- a/.github/workflows/ci-pipeline.yml
+++ b/.github/workflows/ci-pipeline.yml
@@ -382,7 +382,7 @@ jobs:
         dotnet-version: ${{ env.TGS_DOTNET_VERSION }}
 
     - name: Set TGS_TEST_DUMP_API_SPEC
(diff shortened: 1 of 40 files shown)


Cyberboss:
Merging with `[TGSDeploy]`

PR #1772, opened 2024-01-19
Warn when the topic call timeout is too low

Cyberboss:
There is precedence for this

### DIFF
--- a/src/Tgstation.Server.Host/Configuration/GeneralConfiguration.cs
+++ b/src/Tgstation.Server.Host/Configuration/GeneralConfiguration.cs
@@ -187,6 +187,9 @@ namespace Tgstation.Server.Host.Configuration
 			else if (this.GetCopyDirectoryTaskThrottle() < 1)
 				throw new InvalidOperationException(
 					$"{nameof(DeploymentDirectoryCopyTasksPerCore)} is too large for the CPU core count of {Environment.ProcessorCount} and overflows a 32-bit signed integer. Please lower the value!");
+
+			if (ByondTopicTimeout <= 1000)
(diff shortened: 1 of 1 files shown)


ZephyrTFA (approved):

src/Tgstation.Server.Host/Configuration/GeneralConfiguration.cs
```diff
@@ -187,6 +187,9 @@ public void CheckCompatibility(ILogger logger)
 			else if (this.GetCopyDirectoryTaskThrottle() < 1)
 				throw new InvalidOperationException(
 					$"{nameof(DeploymentDirectoryCopyTasksPerCore)} is too large for the CPU core count of {Environment.ProcessorCount} and overflows a 32-bit signed integer. Please lower the value!");
+
+			if (ByondTopicTimeout <= 1000)
+				logger.LogWarning("The timeout for sending BYOND topics is very low ({ms}ms). Topic calls may fail to complete at all!");
```
ZephyrTFA, line 192:
```suggestion
				logger.LogWarning("The timeout for sending BYOND topics is very low ({ms}ms). Topic calls may fail to complete at all!", ByondTopicTimeout);
```

ZephyrTFA (changes requested):

src/Tgstation.Server.Host/Configuration/GeneralConfiguration.cs
Cyberboss: lol holy shit

ZephyrTFA (approved):

PR #1788, opened 2024-02-19
DMAPI Custom Events. Fix potential missing `/home` directory in `.deb`. Standardize DMAPI sleeps. Add support for `file2text` redefinitions in DMAPI. Fix event script priorities. Fix creating invalid `trusted.txt` on Linux

### LINKED ISSUE #1746: Custom event names triggerable from DM code
These should not trigger existing TGS events

Add a DMAPI to trigger a custom event name.

Requested by @MrStonedOne

Cyberboss:
🆑
Fixed event scripts always running with low priority if the configuration option for `Session:LowPriorityDeploymentProcesses` was set instead of only in the deployment pipeline.
Added support for DMAPI triggered custom event scripts. These cannot have the same name as TGS built-in events.
Fixed `.deb` package sometimes not creating a `tgstation-server` home directory.
The nonsense file `$HOME/byond/cfg/trusted.txt` will now no longer be created on Linux.
Fixed the BYOND cache not properly being cleaned on Linux.
/🆑

🆑 Interop API
Added bridge command 7 for triggering custom events. Requires the `eventInvocation` parameter. Consists of `eventName` (string), `parameters` (string array), `notifyCompletion` (bool/int). Response contains the `eventId` string if `notifyCompletion` is set.
Added topic command 12 for notifying about completed custom events. Will only be sent if `notifyCompletion` was set in the triggering bridge request. Contains the `eventId` of the completed event.
/🆑

🆑 DreamMaker API
Changed all instances of `sleep(1)` to `sleep(world.tick_lag)`.
Added support for triggering custom TGS events.
Added support for codebases that override `file2text`.
/🆑

🆑 Web Control Panel
Fixed requiring a branch be specified for quick setup operations.
/🆑

Closes #1746

Merging with `[DMDeploy][APIDeploy][NugetDeploy]`


MrStonedOne:
Just in time for me to remember why i really wanted this, manual shared config sync when i move that to a git(hub) repo.

PR #1791, opened 2024-02-27
v6.3.1: Fix a 500 error. Add a debug message

Cyberboss:
🆑
Fixed a potential 500 error when disabling a user.
/🆑

Merging with `[TGSDeploy][DMDeploy]`


ZephyrTFA:
test 1
test 2
before
after
it works!

PR #1816, opened 2024-05-09
Add Revision Date to V3 DMAPI

### LINKED ISSUE #1815: Provide commit date in V3 DMAPI
Has to be ISO 8601 for compatibility.

Conditionally only works on 514+ due to the timezone parameter in time2text being that new.

ZephyrTFA:
🆑 DreamMaker API
DMAPI v3 now supports revision date information
/🆑

Closes #1815


Cyberboss:
Sub 514 builds failed. Ig you need to wrap it in a #if after all.


Just omit the parameter if it's sub 514.


Also bump the DMAPI patch version in tgs.dm and build/Version.props.

Cyberboss (changes requested):
Sub 514 builds failed. Ig you need to wrap it in a #if after all.

Just omit the parameter if it's sub 514.

Also bump the DMAPI patch version in tgs.dm and build/Version.props.

Cyberboss:
wtf codeql

PR #1882, opened 2024-08-18
Warns when running TGS as root

ZephyrTFA:
🆑
Running as root on Linux now issues a warning.
/🆑

Running any service under the root user when it is not a system service is not recommended. We already install under a dedicated user for the apt packages so we should warn, and in the future error out, when running as root.

### DIFF
--- a/src/Tgstation.Server.Host/Program.cs
+++ b/src/Tgstation.Server.Host/Program.cs
@@ -2,11 +2,13 @@
 using System.Collections.Generic;
 using System.Diagnostics;
 using System.Linq;
+using System.Runtime.InteropServices;
 using System.Text;
 using System.Threading;
 using System.Threading.Tasks;
 
 using Tgstation.Server.Host.Common;
+using Tgstation.Server.Host.Components.Interop.Bridge;
 using Tgstation.Server.Host.Core;
 using Tgstation.Server.Host.Properties;
 using Tgstation.Server.Host.System;
@@ -65,6 +67,34 @@ namespace Tgstation.Server.Host
 					Debugger.Launch();
 
 				args = listArgs.ToArray();
+			}
+
+			if(RuntimeInformation.IsOSPlatform(OSPlatform.Linux))
+			{
+				var proc = new Process
+				{
+    				StartInfo = new ProcessStartInfo
+    				{
+        				FileName = "id",
+       					Arguments = "-u",
+        				UseShellExecute = false,
+        				RedirectStandardOutput = true,
+       					CreateNoWindow = true
+    				}
+				};
+				proc.Start();
+				await proc.WaitForExitAsync();
+				if(proc.ExitCode is not 0 || !int.TryParse(await proc.StandardOutput.ReadToEndAsync(), out var uid)) {
+					Console.Error.WriteLine("Failed to obtain user id.");
+					Environment.Exit(1);
+					return;
+				}
+				if(uid is 0)
+				{
+					Console.Error.WriteLine("TGS is being run as root. This is not recommended and will prevent launching in a future version!");
+					// Environment.Exit(1);
+					// return;
+				}
 			}
 
 			var program = new Program();


src/Tgstation.Server.Host/Program.cs
```diff
@@ -67,6 +70,34 @@ public static async Task<int> Main(string[] args)
...
+				await proc.WaitForExitAsync();
+				if (proc.ExitCode is not 0 || !int.TryParse(await proc.StandardOutput.ReadToEndAsync(), out var uid))
+				{
+					Console.Error.WriteLine("Failed to obtain user id.");
+					return 1;
+				}
+
+				if (uid is 0)
+				{
+					Console.Error.WriteLine("TGS is being run as root. This is not recommended and will prevent launching in a future version!");
```
ZephyrTFA, line 97:
```suggestion
					Console.Error.WriteLine("TGS is being run as root. This is not recommended and may prevent launch in a future version.");
```

src/Tgstation.Server.Host/Program.cs
```diff
@@ -67,6 +70,34 @@ public static async Task<int> Main(string[] args)
...
+				if (proc.ExitCode is not 0 || !int.TryParse(await proc.StandardOutput.ReadToEndAsync(), out var uid))
+				{
+					Console.Error.WriteLine("Failed to obtain user id.");
+					return 1;
+				}
+
+				if (uid is 0)
+				{
+					Console.Error.WriteLine("TGS is being run as root. This is not recommended and may prevent launch in a future version.");
+				}
```
AffectedArc07, line 98: Fyi you can do `Environment.UserName` and just see if thats `root` or not.

Up to you if you want to check on `uid==0`, or prevent the need to spawn extra processes.

ZephyrTFA: Not a good solution as you can change the name of root

src/Tgstation.Server.Host/Program.cs
```diff
@@ -67,6 +70,34 @@ public static async Task<int> Main(string[] args)
...
+				await proc.WaitForExitAsync();
+				if (proc.ExitCode is not 0 || !int.TryParse(await proc.StandardOutput.ReadToEndAsync(), out var uid))
+				{
+					Console.Error.WriteLine("Failed to obtain user id.");
+					return 1;
+				}
+
+				if (uid is 0)
+				{
+					Console.Error.WriteLine("TGS is being run as root. This is not recommended and may prevent launch in a future version.");
```
Cyberboss, line 97: This is not the place to put arbitrary startup checks. There's a function called `CheckSystemCompatibility` somewhere. Use that instead.

src/Tgstation.Server.Host/Program.cs
```diff
@@ -67,6 +70,34 @@ public static async Task<int> Main(string[] args)
 				args = listArgs.ToArray();
 			}
 
+			if (InteropServices.RuntimeInformation.IsOSPlatform(InteropServices.OSPlatform.Linux))
+			{
+				using var proc = new Process
```
Cyberboss, line 75: This is unnecessary. There should be a getuid function available in the `Syscall` class somewhere.

src/Tgstation.Server.Host/Program.cs
Cyberboss: Should also be "as the root account".

Cyberboss: These need to use `ILogger.LogWarning` instead of direct console output.

src/Tgstation.Server.Host/Program.cs
Cyberboss: And platform specific checks should be abstracted, implement it as an `IsRoot` property of `ISystemIdentity`.

src/Tgstation.Server.Host/Program.cs
```diff
@@ -67,6 +70,34 @@ public static async Task<int> Main(string[] args)
 				args = listArgs.ToArray();
 			}
 
+			if (InteropServices.RuntimeInformation.IsOSPlatform(InteropServices.OSPlatform.Linux))
```
Cyberboss, line 73: Needs to use `!(IPIatformIdentifier.IsWindows)`

Cyberboss (changes requested):

Cyberboss:
Also, ngl, these webedits really make me want to go against my principle of only using merge commits.

ZephyrTFA:
> Also, ngl, these webedits really make me want to go against my principle of only using merge commits.

not a single commit was a web edit; it was me struggling to get dotnet to actually build locally until I realized that I had to manually specify NoWindows as a configuration because it is not automatically detected

src/Tgstation.Server.Host/Components/InstanceManager.cs
```diff
@@ -675,6 +675,11 @@ void CheckSystemCompatibility()
 			{
 				if (!systemIdentity.CanCreateSymlinks)
 					throw new InvalidOperationException($"The user running {Constants.CanonicalPackageName} cannot create symlinks! Please try running as an administrative user!");
+
+				if (!platformIdentifier.IsWindows && systemIdentity.IsSuperUser)
+				{
+					logger.LogWarning("TGS is being run as the root account. This is not recommended and may prevent launch in a future version.");
```
Cyberboss, line 681: This is never really gonna happen
```suggestion
					logger.LogWarning("TGS is being run as the root account. This is not recommended.");
```

src/Tgstation.Server.Host/Security/WindowsSystemIdentity.cs
```diff
@@ -22,7 +22,10 @@ sealed class WindowsSystemIdentity : ISystemIdentity
 		public string Username => userPrincipal?.Name ?? identity!.Name;
 
 		/// <inheritdoc />
-		public bool CanCreateSymlinks => canCreateSymlinks ?? throw new NotSupportedException();
+		public bool CanCreateSymlinks => IsSuperUser;
+
+		/// <inheritdoc />
+		public bool IsSuperUser => IsSuperUser;
```
Cyberboss, line 28: Infinite recursion!
```suggestion
		public bool IsSuperUser => isAdmin ?? throw new NotSupportedException();
```

src/Tgstation.Server.Host/Program.cs
```diff
@@ -11,6 +11,8 @@
 using Tgstation.Server.Host.Properties;
 using Tgstation.Server.Host.System;
 
+using Process = System.Diagnostics.Process;
```
Cyberboss, line 14: Can you undo the changes to this file please?

Cyberboss (changes requested):

src/Tgstation.Server.Host/Program.cs
```diff
@@ -11,6 +11,8 @@
...
+
```
ZephyrTFA, line 15:
```suggestion
```

src/Tgstation.Server.Host/Program.cs
```diff
@@ -40,7 +42,7 @@ public Program()
 		/// Entrypoint for the <see cref="Program"/>.
 		/// </summary>
 		/// <param name="args">The command line arguments.</param>
-		/// <returns>A <see cref="Task"/> resulting in the <see cref="global::System.Diagnostics.Process.ExitCode"/>.</returns>
+		/// <returns>A <see cref="Task"/> resulting in the <see cref="Process.ExitCode"/>.</returns>
```
ZephyrTFA, line 45:
```suggestion
		/// <returns>A <see cref="Task"/> resulting in the <see cref="[Process.ExitCode](global::System.Diagnostics.Process.ExitCode)"/>.</returns>
```

Cyberboss (approved):

src/Tgstation.Server.Host/Components/InstanceManager.cs
```diff
@@ -675,6 +675,11 @@ void CheckSystemCompatibility()
 			{
 				if (!systemIdentity.CanCreateSymlinks)
 					throw new InvalidOperationException($"The user running {Constants.CanonicalPackageName} cannot create symlinks! Please try running as an administrative user!");
+
+				if (!platformIdentifier.IsWindows && systemIdentity.IsSuperUser)
```
Cyberboss, line 679: Sorry to keep coming back to this, but I forgot this will always fire on docker.

Please create an `InternalConfiguration.RunningInDocker` setting and make sure the docker image has it set by adding `ENV Internal__RunningInDocker=true` near the end of `build/Dockerfile`. The warning should not fire if it's `true`.

`InstanceManager` doesn't have a reference to the internal config yet so you'll also have to add one. It's the same as the already present `swarmConfiguration` field.

Cyberboss (changes requested):

src/Tgstation.Server.Host/Configuration/InternalConfiguration.cs
```diff
@@ -25,6 +27,11 @@ public sealed class InternalConfiguration
 		/// </summary>
 		public bool UsingSystemD { get; set; }
 
+		/// <summary>
+		/// If the server is running inside of a Docker container.
+		/// </summary>
+		public bool UsingDocker => Environment.GetEnvironmentVariable("Internal__UsingDocker")?.Equals("true", StringComparison.OrdinalIgnoreCase) ?? false;
```
Cyberboss, line 33: It needs to be just `{ get; set; }`, config system handles the rest.

Cyberboss (changes requested):

Cyberboss (approved):

PR #1919, opened 2024-09-03
Fix client caching with app installation tokens

Cyberboss:
:cl:
Fixed issue where GitHub app tokens were being cached longer than their 1 hour expiry.
/:cl:


Cyberboss:
Rather not broadcast how version reporting is implemented.

PR #1974, opened 2024-10-15
life support for v3

ZephyrTFA:
I cannot test this locally because WiX refuses to install and link on my local repository.


Cyberboss:
The V3 branch is frozen because I can't figure out how to deploy it anymore

PR #2174, opened 2025-03-20
v6.15.2: Fix `X-Forwarded-*` headers again

Cyberboss:
AHHHHHHHHHHHHHHHHH

🆑
Fixed `X-Forwarded-*` headers only working for localhost reverse proxies.
Guard against rogue database OAuthConnection rows being created.
/🆑


Crossedfall (approved):
Works in the pre-release builds 👏

PR #2223, opened 2025-05-06
v6.18.0: Send a chat message when auto-updates fail due to testmerge conflicts

Cyberboss:
🆑
A deployment chat message will now be sent during the instance auto-update process if it failed due to a conflicting test merge.
/🆑


optimumtact:
this is pointless noise that will not actually result in any meaningful outcome.

PR #2325, opened 2025-08-12
Convert all POSTs to PUTs

Drulikar:

🆑 REST API
All REST API methods using PUT are replaced with POST. This also requires other POSTs to use a different route.
/🆑

This has been to resolve a longstanding, intermittent issue where actions that are performed using a PUT request will fail (`X-Powered-By: tgstation-server` header is not present). For me on an AT&T ISP routing to CM which uses cloudflare proxying and NGINX this *usually* isn't an issue but I have periodically been able to reproduce the problem. But for other people, even people in the US, its the opposite case where almost never will a PUT request succeed.

As far as why, I haven't been able to research definitive reasons why, but for example see discussion here https://www.reddit.com/r/ProgrammerTIL/comments/4pztl8/other_the_real_difference_between_http_verbs_put/ 
> POST works every time. Use anything but GET and POST, and you risk running into something that doesn't support it. A framework, some kind of proxy or "firewall"

I have been testing a version of this for our production TGS for CM PVE (see commit https://github.com/tgstation/tgstation-server/commit/aba3f21f495ae2b3ee0b0c6700973203f10411ec for some info of how I was testing it) and it has eliminated all issues with our GMs being unable to perform actions in TGS.

Let me know if you want a different version number, there's anything I missed (not completely clear what if anything else needs to change for the github pages version of the API).


Cyberboss (changes requested):
So one major thing, this should be "All PUT REST API endpoints now have POST alternatives" so we don't have to make a breaking change

build/Version.props
```diff
@@ -5,7 +5,7 @@
   <PropertyGroup>
     <TgsCoreVersion>6.19.0</TgsCoreVersion>
     <TgsConfigVersion>5.8.0</TgsConfigVersion>
-    <TgsRestVersion>10.14.0</TgsRestVersion>
+    <TgsRestVersion>10.15.0</TgsRestVersion>
```
Cyberboss, line 8:
```suggestion
    <TgsRestVersion>10.14.0</TgsRestVersion>
```
10.14 still hasn't released

Cyberboss (changes requested):

src/Tgstation.Server.Client/ApiClient.cs
```diff
@@ -522,6 +522,9 @@ protected virtual async ValueTask<TResult> RunRequest<TResult>(
 					if (fileDownload)
 						request.Headers.Accept.Add(new MediaTypeWithQualityHeaderValue(MediaTypeNames.Application.Octet));
 
+					if (method == HttpMethod.Put)
+						request.Headers.Warning.Add(new WarningHeaderValue(299, "-", "HttpMethod.Put is depreciated and will be replaced with HttpMethod.Post in next major release."));
+
```
Cyberboss, line 527: These are client headers. They shouldn't care about the deprecation
```suggestion
```

src/Tgstation.Server.Client/IApiClient.cs
```diff
@@ -53,7 +53,7 @@ ValueTask<IAsyncDisposable> CreateHubConnection<THubImplementation>(
 			where THubImplementation : class;
 
 		/// <summary>
-		/// Run an HTTP PUT request.
+		/// Run an HTTP POST request.
```
Cyberboss, line 56: The `Create` and `Delete` methods correspond to HTTP verbs. They should be removed entirely and replaced with their `Update` variants where they were used.

Also did you not want to remove the HTTP PATCH verbs in this change? You seem to have neglected it.

src/Tgstation.Server.Host/Extensions/ApplicationBuilderExtensions.cs
```diff
@@ -161,6 +162,13 @@ public static void UseServerBranding(this IApplicationBuilder applicationBuilder
 			applicationBuilder.Use((context, next) =>
 			{
 				context.Response.Headers.Add("X-Powered-By", assemblyInformationProvider.VersionPrefix);
+				if (context.Request.Method == HttpMethods.Put)
+				{
+					var logger = GetLogger(context);
+					context.Response.Headers.Add(HeaderNames.Warning, "299 - HttpMethod.Put is depreciated and will be replaced with HttpMethod.Post in next major release.");
```
Cyberboss, line 168: Remove the reference to HTTP 299 as that's not going to affect the status code at all and its deprecated anyway.

Also `HttpMethod` is an ASP NET core implementation detail

```suggestion
					context.Response.Headers.Add(HeaderNames.Warning, "HTTP PATCH/PUT/DELETE verbs are depreciated and will be replaced with POST with updated endpoints in an upcoming major REST API release.");
```

src/Tgstation.Server.Host/Extensions/ApplicationBuilderExtensions.cs
```diff
@@ -161,6 +162,13 @@ public static void UseServerBranding(this IApplicationBuilder applicationBuilder
 			applicationBuilder.Use((context, next) =>
 			{
 				context.Response.Headers.Add("X-Powered-By", assemblyInformationProvider.VersionPrefix);
+				if (context.Request.Method == HttpMethods.Put)
```
Cyberboss, line 165: This should be done in `ApiController.HookExecuteAction` not here. It's not what the function is documented to do and the scope of requests that go through here affect all HTTP requests, not just the REST API.

src/Tgstation.Server.Api/Routes.cs
```diff
@@ -112,6 +112,26 @@ public static class Routes
...
+		/// <summary>
+		/// The postfix for launch operations.
+		/// </summary>
+		public const string Launch = "Launch";
+
+		/// <summary>
+		/// The postfix for update operations.
+		/// </summary>
+		public const string Update = "Update";
+
```
Cyberboss, line 129: We shouldn't change the path of existing POST endpoints if it's not necessary

```suggestion
```

src/Tgstation.Server.Host/Controllers/ChatController.cs
```diff
@@ -269,6 +270,7 @@ public async ValueTask<IActionResult> GetId(long id, CancellationToken cancellat
 		/// <response code="204">Update applied successfully. <see cref="ChatBot"/> not returned based on user permissions.</response>
 		/// <response code="410">The <see cref="ChatBot"/> with the given ID does not exist in this instance.</response>
 		[HttpPost]
+		[HttpPost(Routes.Update)]
```
Cyberboss, line 273: We shouldn't change the path of existing POST endpoints if it's not necessary

```suggestion
```

src/Tgstation.Server.Host/Controllers/ConfigurationController.cs
```diff
@@ -68,6 +68,7 @@ public ConfigurationController(
 		/// <response code="200">File updated successfully.</response>
 		/// <response code="202">File upload ticket created successfully.</response>
 		[HttpPost]
+		[HttpPost(Routes.Update)]
```
Cyberboss, line 71: We shouldn't change the path of existing POST endpoints if it's not necessary

```suggestion
```

src/Tgstation.Server.Host/Controllers/InstanceController.cs
```diff
@@ -342,6 +343,7 @@ await DatabaseContext
 		/// <response code="202">Instance updated successfully and relocation job created.</response>
 		/// <response code="410">The database entity for the requested instance could not be retrieved. The instance was likely detached.</response>
 		[HttpPost]
+		[HttpPost(Routes.Update)]
```
Cyberboss, line 346: We shouldn't change the path of existing POST endpoints if it's not necessary

```suggestion
```

src/Tgstation.Server.Host/Controllers/InstancePermissionSetController.cs
```diff
@@ -136,6 +137,7 @@ public async ValueTask<IActionResult> Create([FromBody] InstancePermissionSetReq
 		/// <response code="200"><see cref="InstancePermissionSet"/> updated successfully.</response>
 		/// <response code="410">The requested <see cref="InstancePermissionSet"/> does not currently exist.</response>
 		[HttpPost]
+		[HttpPost(Routes.Update)]
```
Cyberboss, line 140: We shouldn't change the path of existing POST endpoints if it's not necessary

```suggestion
```

src/Tgstation.Server.Host/Controllers/RepositoryController.cs
```diff
@@ -344,6 +345,7 @@ public async ValueTask<IActionResult> Read(CancellationToken cancellationToken)
 		/// <response code="202">Updated the repository settings successfully and a <see cref="JobResponse"/> was created to make the requested git changes.</response>
 		/// <response code="410">The database entity for the requested instance could not be retrieved. The instance was likely detached.</response>
 		[HttpPost]
+		[HttpPost(Routes.Update)]
```
Cyberboss, line 348: We shouldn't change the path of existing POST endpoints if it's not necessary

```suggestion
```

src/Tgstation.Server.Host/Controllers/UserController.cs
```diff
@@ -81,6 +82,7 @@ public ValueTask<IActionResult> Create([FromBody] UserCreateRequest model, Cance
 		/// <response code="404">Requested <see cref="EntityId.Id"/> does not exist.</response>
 		/// <response code="410">Requested <see cref="Api.Models.Internal.UserApiBase.Group"/> does not exist.</response>
 		[HttpPost]
+		[HttpPost(Routes.Update)]
```
Cyberboss, line 85: We shouldn't change the path of existing POST endpoints if it's not necessary

```suggestion
```

src/Tgstation.Server.Host/Controllers/UserGroupController.cs
```diff
@@ -103,6 +104,7 @@ public async ValueTask<IActionResult> Create([FromBody] UserGroupCreateRequest m
 		/// <response code="200"><see cref="UserGroup"/> updated successfully.</response>
 		/// <response code="410">The requested <see cref="UserGroup"/> does not currently exist.</response>
 		[HttpPost]
+		[HttpPost(Routes.Update)]
```
Cyberboss, line 107: We shouldn't change the path of existing POST endpoints if it's not necessary

```suggestion
```

Cyberboss (changes requested):

src/Tgstation.Server.Api/Routes.cs
Drulikar: I think it makes the api cleaner and more obvious whats being done, but I'm exhausted w/ this PR.

src/Tgstation.Server.Client/IApiClient.cs
Drulikar: PR set out to remove only PUT since I know for sure PUT is unreliable. I have no information or first hand knowledge of DELETE or PATCH. But I guess I can see if I can do it too...

Drulikar:
I am still going to move forward with this PR. How it's done here is non breaking so you shouldn't care. I will be making a PR to fully remove them on a different branch though because the API does not need to be designed in this way and done in an unreliable way. If you have software that can be hardened to user error like bad infrastructure then you should do so like this is doing. I have asked for months what people think is the "infrastructure issue" and no one has any ideas on what to explore. But I do know this works.

I only linked an old reddit post because as I said I can't find definitive documentation on why PUT is an unreliable verb compared to POST. 

If this PR is refused I will just stop being a contributor here and fork my own version of tgs so proper changes like this can be done.

alexkar598:
> I am still going to move forward with this PR. How it's done here is non breaking so you shouldn't care. I will be making a PR to fully remove them on a different branch though because the API does not need to be designed in this way and done in an unreliable way. 

The way is not unreliable, This is standard REST API practice. Also I care about the API not being made worse even if the making it worse is done is a backwards compatible way.

> If you have software that can be hardened to user error like bad infrastructure then you should do so like this is doing. I have asked for months what people think is the "infrastructure issue" and no one has any ideas on what to explore. But I do know this works.

I literally pointed you towards what you should debug 2 months ago and you were being a dick to me so I gave up. Lobotomizing the API is not "hardening" in the same way forcing a downgrade to HTTP/1.1 isn't hardening. 

> I only linked an old reddit post because as I said I can't find definitive documentation on why PUT is an unreliable verb compared to POST.

Because PUT is not an unreliable verb compared to POST. Your failure to produce documentation is a testament to this.

> If this PR is refused I will just stop being a contributor here and fork my own version of tgs so proper changes like this can be done.

???

> If you rather things be routed more literally to indicate what they do (e.g. now the /{id} routes that could instead be /Delete/{id} or something), then by all means suggest changes but you also will need to convince Dominion to not RC reverting so.

No clue what RC reverting but /delete/{id} is bad API design and should instead use semantic HTTP verbs to be more explicit with the behavior of the API.




Frankly if you want to lobotomize the api to hack around things, you're better off configuring nginx to do all the rewrites you want in the world

PR #2441, opened 2026-08-15
Add Discord slash command support

### LINKED ISSUE #2437: Get rid of the Message Content Intent
**Is your feature request related to a problem? Please describe.**

Discord is changing the rules for privileged intents

Starting 10 June 2026, apps visible to 10000 or more users will need to apply for privileged intent access and repeat the review every year

TGS currently uses the Message Content intent to read messages and handle commands such as `!tgs`. The same path is also used for commands registered through DMAPI

Discord now recommends moving bots from text commands to slash commands instead of relying on the Message Content intent

Image

Links:

* https://support-dev.discord.com/hc/en-us/articles/40281523410967-Changes-to-Privileged-Intent-Access-for-Discord-Apps
* https://docs.discord.com/developers/gateway/getting-started-with-privileged-intent-review

**Describe the solution you'd like**

Add a single discord slash command with a free form text argument, for example:

```text
/tgs: <foobar>
/tgs: info
/tgs: adminwho
```

TGS could pass the supplied text into the existing command parser as if the user had sent:

```text
!tgs <foobar>
!tgs info
!tgs adminwho
```

This would avoid registering every TGS or DMAPI command separately in discord and should allow the bot to work without the Message Content intent

The slash command should still keep the user, guild and channel context needed by the current command system

For commands that take longer to complete, the bot can defer the discord interaction response and send the result when it is ready

I think the existing `!tgs` message command maybe could remain as an optional compatibility mode

**Describe alternatives you've considered**

* Keep using `!tgs` and apply for Message Content intent access. This adds a review process and yearly reapproval for larger apps. Huge pain in the ass, and there is good reason to believe discord may simply refuse to approve the verification
* Register every TGS and DMAPI command as its own discord slash command. This would be more native to discord, but dynamic DMAPI commands would need to be kept in sync
* Remove incoming discord commands and keep discord only for notifications. This avoids the intent requirement, but removes existing functionality

PetyaFX:
[Base Branch]: dev

🆑 Core
Added a global Discord slash command for TGS chat bot commands.
Added autocomplete for built-in and active DMAPI chat command names in the slash command's required command field.
Added an optional arguments field to the Discord slash command.
/🆑

🆑 Configuration
Added `General:DiscordSlashCommandName` to configure the Discord slash command name, defaulting to `tgs`.
/🆑

[Why]:

Closes #2437

Discord's Message Content Intent is privileged and may be unavailable preventing normal message commands such as `!tgs` and bot mentions from being used. The slash command provides a supported command interface without requiring that intent, while autocomplete makes the available built-in and active DMAPI commands discoverable.

image
Screenshot 2026-08-15 191809
Screenshot 2026-08-15 191823

### DIFF
--- a/README.md
+++ b/README.md
@@ -269,6 +269,8 @@ Create an `appsettings.Production.yml` file next to `appsettings.yml`. This will
 
 - `General:GitHubAccessToken`: Specify a classic GitHub personal access token with no scopes here to highly mitigate the possiblity of 429 response codes from GitHub requests
 
+- `General:DiscordSlashCommandName`: The Discord slash command name for TGS chat commands. Defaults to `tgs`; use 1-32 letters, digits, `-`, `_`, or apostrophes, without the leading `/`. Restart is required.
+
 - `General:SkipAddingByondFirewallException`: Set to `true` if you have Windows firewall disabled
 
 - `General:AdditionalEventScriptsDirectories`: An array of directories that are considered to contain EventScripts alongside instance directories. Working directory for executed scripts will remain the instance EventScripts directory.
@@ -624,11 +626,20 @@ TGS supports creating infinite chat bots for notifying staff or players of thing
 
 - Internet Relay Chat (IRC)
 - Discord
-  - NOTE: Bot MUST have Message Content Intent enabled
+  - The Message Content Intent enables `!tgs` and mention commands in guild channels.
+  - Without the intent, use the `/tgs` slash command in guild channels. Its required **command** argument autocompletes built-in and active DMAPI command names; **arguments** is optional.
+  - Set `General:DiscordSlashCommandName` to change the slash command name from its default, `/tgs`.
+  - To use Discord without the Message Content Intent, configure the bot in the Developer Portal:
+    1. Open the app and go to **OAuth2 - URL Generator**.
+    2. Select the `bot` and `applications.commands` scopes.
+    3. Grant the bot at least these channel permissions already used by TGS:
+       - `View Channels`
+       - `Send Messages`
+       - `Read Message History` for normal reply/history behavior.
 
 More can be added by providing a new implementation of the [IProvider](src/Tgstation.Server.Host/Components/Chat/Providers/IProvider.cs) interface
 
-Bots have a set of built-in commands that can be triggered via `!tgs`, mentioning, or private messaging them. Along with these, custom commands can be defined using the DMAPI by creating a subtype of the `/datum/tgs_chat_command` type (See `tgs.dm` for details). Invocation for custom commands can be restricted to certain channels.
+Bots have a set of built-in commands that can be triggered via `!tgs`, `/tgs`, or mentioning them. Along with these, custom commands can be defined using the DMAPI by creating a subtype of the `/datum/tgs_chat_command` type (See `tgs.dm` for details). Invocation for custom commands can be restricted to certain channels.
 
 #### Static Files
 
--- a/src/Tgstation.Server.Host/Components/Chat/ChatManager.cs
+++ b/src/Tgstation.Server.Host/Components/Chat/ChatManager.cs
@@ -297,7 +297,7 @@ namespace Tgstation.Server.Host.Components.Chat
 					disconnectTask = Task.CompletedTask;
 				if (newSettingsEnabled)
 				{
-					provider = providerFactory.CreateProvider(newSettings);
+					provider = providerFactory.CreateProvider(newSettings, GetCommandNames);
 					providers.Add(newSettingsId, provider);
 				}
 			}
@@ -593,6 +593,26 @@ namespace Tgstation.Server.Host.Components.Chat
 					Text = message,
 				},
 				cancellationToken);
+		}
+
+		/// <summary>
+		/// Gets the currently available chat command names.
+		/// </summary>
+		/// <returns>The command names.</returns>
+		IReadOnlyList<string> GetCommandNames()
+		{
+			var commands = new List<string> { "help" };
+			commands.AddRange(builtinCommands.Values.Select(command => command.Name));
+			lock (trackingContexts)
+				commands.AddRange(
+					trackingContexts
+						.Where(trackingContext => trackingContext.Active)
+						.SelectMany(trackingContext => trackingContext.CustomCommands.Select(command => command.Name)));
+
+			return commands
+				.Distinct(StringComparer.OrdinalIgnoreCase)
+				.OrderBy(command => command, StringComparer.OrdinalIgnoreCase)
+				.ToList();
 		}
 
 		/// <summary>
--- a/src/Tgstation.Server.Host/Components/Chat/Providers/DiscordProvider.cs
+++ b/src/Tgstation.Server.Host/Components/Chat/Providers/DiscordProvider.cs
@@ -11,6 +11,8 @@ using Microsoft.Extensions.DependencyInjection;
 using Microsoft.Extensions.Logging;
 using Microsoft.Extensions.Options;
 
+using OneOf;
+
 using Remora.Discord.API.Abstractions.Gateway.Commands;
 using Remora.Discord.API.Abstractions.Gateway.Events;
 using Remora.Discord.API.Abstractions.Objects;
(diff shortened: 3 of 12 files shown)


Cyberboss:
I need to fix CI, sorry. Will get to that when I can. Looks okay from a conceptual standpoint

Cyberboss:
Sorry for the delays on this. Been super busy with life/work and still repairing CI.

This definitely needs to go in though for various reasons. Will get to it

README.md
```diff
@@ -624,11 +626,20 @@ TGS supports creating infinite chat bots for notifying staff or players of thing
...
 - Internet Relay Chat (IRC)
 - Discord
-  - NOTE: Bot MUST have Message Content Intent enabled
+  - The Message Content Intent enables `!tgs` and mention commands in guild channels.
+  - Without the intent, use the `/tgs` slash command in guild channels. Its required **command** argument autocompletes built-in and active DMAPI command names; **arguments** is optional.
+  - Set `General:DiscordSlashCommandName` to change the slash command name from its default, `/tgs`.
+  - To use Discord without the Message Content Intent, configure the bot in the Developer Portal:
+    1. Open the app and go to **OAuth2 - URL Generator**.
+    2. Select the `bot` and `applications.commands` scopes.
+    3. Grant the bot at least these channel permissions already used by TGS:
```
Cyberboss, line 635:
```suggestion
    1. Select the `bot` and `applications.commands` scopes.
    1. Grant the bot at least these channel permissions already used by TGS:
```

src/Tgstation.Server.Host/Configuration/GeneralConfiguration.cs
```diff
@@ -132,6 +132,11 @@ public sealed class GeneralConfiguration : ServerInformationBase
 		/// </summary>
 		public bool UseBasicWatchdog { get; set; }
 
+		/// <summary>
+		/// The Discord slash command name for TGS chat commands.
+		/// </summary>
+		public string DiscordSlashCommandName { get; set; } = "tgs";
```
Cyberboss, line 138: Use a common `const`ant for this please.

src/Tgstation.Server.Host/Components/Chat/ChatManager.cs
```diff
@@ -595,6 +595,26 @@ public ValueTask HandleRestart(Version? updateVersion, bool handlerMayDelayShutd
 				cancellationToken);
 		}
 
+		/// <summary>
+		/// Gets the currently available chat command names.
+		/// </summary>
+		/// <returns>The command names.</returns>
+		IReadOnlyList<string> GetCommandNames()
+		{
+			var commands = new List<string> { "help" };
```
Cyberboss, line 604: Same here, I'm pretty sure this string exists somewhere

src/Tgstation.Server.Host/Components/Chat/Providers/IProviderFactory.cs
```diff
@@ -11,7 +14,8 @@ interface IProviderFactory
 		/// Create a <see cref="IProvider"/>.
 		/// </summary>
 		/// <param name="settings">The <see cref="ChatBot"/> containing settings for the new provider.</param>
+		/// <param name="commandNamesFactory">Gets the currently available chat command names.</param>
 		/// <returns>A new <see cref="IProvider"/>.</returns>
-		IProvider CreateProvider(ChatBot settings);
+		IProvider CreateProvider(ChatBot settings, Func<IReadOnlyList<string>> commandNamesFactory);
```
Cyberboss, line 19:
```suggestion
		IProvider CreateProvider(ChatBot settings, Func<IEnumerable<string>> commandNamesFactory);
```

src/Tgstation.Server.Host/Components/Chat/Providers/DiscordProvider.cs
```diff
@@ -898,6 +1287,93 @@ async ValueTask<IEnumerable<IChannel>> GetGuildChannels(IPartialGuild guild)
...
+				? commandData.Options.Value.FirstOrDefault(option => option.Name.Equals(SlashCommandOptionName, StringComparison.Ordinal) && option.IsFocused.HasValue && option.IsFocused.Value)
+				: null;
+			var focusedValue = focusedCommand != null && focusedCommand.Value.HasValue && focusedCommand.Value.Value.IsT0
+				? focusedCommand.Value.Value.AsT0
+				: String.Empty;
+
+			var choices = commandNamesFactory()
+				.Where(command => String.IsNullOrEmpty(focusedValue) || command.StartsWith(focusedValue, StringComparison.OrdinalIgnoreCase))
+				.Take(MaxAutocompleteChoices)
+				.Select(command => (IApplicationCommandOptionChoice)new ApplicationCommandOptionChoice(command, OneOf<string, int, double>.FromT0(command)))
```
Cyberboss, line 1314: Do you know what the interaction is if multiple TGS Discord bots are in the same channel?

src/Tgstation.Server.Host/Components/Chat/Providers/DiscordProvider.cs
```diff
@@ -56,6 +58,109 @@ public override string BotMention
...
+			=> applicationFlags.HasFlag(ApplicationFlags.GatewayMessageContent)
+				|| applicationFlags.HasFlag(ApplicationFlags.GatewayMessageContentLimited);
+
+		/// <summary>
+		/// Checks if an application command is a registered TGS slash command that should be removed from a stale scope.
+		/// </summary>
+		/// <param name="command">The application command.</param>
+		/// <param name="configuredCommandName">The configured slash command name.</param>
+		/// <returns><see langword="true"/> if the command is stale.</returns>
+		internal static bool IsTgsSlashCommand(IApplicationCommand command, string configuredCommandName)
```
Cyberboss, line 136: Do these need to be internal/public? I don't see where they're used outside this class.

If it's for tests, prefer using `public` since the class itself is already internal

src/Tgstation.Server.Host/Components/Chat/Providers/DiscordProvider.cs
```diff
@@ -141,19 +278,48 @@ public override string BotMention
...
+		/// </summary>
+		/// <param name="jobManager">The <see cref="IJobManager"/> for the <see cref="Provider"/>.</param>
+		/// <param name="asyncDelayer">The <see cref="IAsyncDelayer"/> for the <see cref="Provider"/>.</param>
+		/// <param name="logger">The <see cref="ILogger"/> for the <see cref="Provider"/>.</param>
+		/// <param name="assemblyInformationProvider">The value of <see cref="assemblyInformationProvider"/>.</param>
+		/// <param name="generalConfigurationOptions">The value of <see cref="generalConfigurationOptions"/>.</param>
+		/// <param name="chatBot">The <see cref="ChatBot"/> for the <see cref="Provider"/>.</param>
+		/// <param name="commandNamesFactory">Gets the currently available chat command names.</param>
+		/// <param name="serviceProviderOverride">The optional service provider override for tests.</param>
+		internal DiscordProvider(
```
Cyberboss, line 307: private if it's not used by tests, public otherwise

Cyberboss (changes requested):
Branch should compile now, take care of these and we can run it:

src/Tgstation.Server.Host/Components/Chat/ChatManager.cs
PetyaFX: Only found literals, added

src/Tgstation.Server.Host/Components/Chat/Providers/DiscordProvider.cs
PetyaFX: Each bot registers the command under its own app, so discord lists `/tgs` per bot and delivers the interaction only to the bot whose command was picked, so no cross talk I guess

PetyaFX:
These are flaky I assume? There are some file locks in the artifact

Cyberboss:
Yeah, the live tests group is flaky. Would have authored them differently now, but that was 5 years ago.

There are automatic retries, and if those fail I'll get it

Cyberboss:
As long as they all don't fail simultaneously, it's probably not from your change

Cyberboss (changes requested):
Sorry, I forgot that you need to bump the `TgsConfigVersion` in `build/Version.props` and add that note to your changelist under a `**Configuration**` section.

Looks fine otherwise.

